PR #1, opened 2017-12-17
Blur

davezuch:


src/Utils.purs
```diff
@@ -36,21 +36,39 @@ inContainer :: ∀ item t f. (Unit -> t Unit) -> f -> Container.Query item t f
...
+  ]
+
+getContainerProps = augmentHTML
+  [ HE.onMouseDown $ HE.input_ $ Container.Mouse Container.Down
+  , HE.onMouseUp   $ HE.input_ $ Container.Mouse Container.Up
+  , HE.onBlur      $ HE.input_ $ Container.Blur
+  , HP.tabIndex 0
+  ]
+
+getChildProps = augmentHTML
```
thomashoneyman, line 63: @whoadave Is this something that we would provide to end users of the lib? Or is this an example of something they'd write custom to add props to something they're putting into the render function?

src/Utils.purs
```diff
@@ -36,21 +36,39 @@ inContainer :: ∀ item t f. (Unit -> t Unit) -> f -> Container.Query item t f
 inContainer = Container.ParentQuery <<< H.action
 
 getInputProps = augmentHTML
```
thomashoneyman: @whoadave How would this work if we want to close the menu after a single selection? Right now, it keeps the menu open even if you click in on a child item. If the user wanted to instead have the menu close after any selection is made, what would that look like on their end (with only access to the render function)?

src/Primitive/Container.purs
```diff
@@ -140,6 +155,23 @@ component render =
...
+        st <- H.get
+        if not st.open then pure a else a <$ case ms of
+          Down -> do
+            H.liftAff $ log $ "mouse: down"
+            H.modify (_ { mouseDown = true })
+          Up -> do
+            H.liftAff $ log $ "mouse: up"
+            H.modify (_ { mouseDown = false })
+
+      Blur a -> do
```
thomashoneyman, line 168: @whoadave How would this work if we want to close the menu after a single selection? Right now, it keeps the menu open even if you click in on a child item. If the user wanted to instead have the menu close after any selection is made, what would that look like on their end (with only access to the render function)?

src/Utils.purs
davezuch: This is something we provide to users of the lib. It needs to be used for any elements rendered inside the container that can steal away focus from the toggle or item elements e.g. custom buttons, links, form controls in a dropdown

If you build the example and run it, you can see how it's applied to the "Click Me" button. Without the childProps, clicking on the button and then clicking away would not hide the dropdown as expected

src/Primitive/Container.purs
davezuch: The parent would have to send the action back down to the child, I added the following to line 121 of the example `Container.purs` to get that to work:

`_ <- H.query unit $ H.action $ Container.Visibility Container.Off`

thomashoneyman:
Cool. I'd like to have an example that displays how each variant could be done (selected items stay in list vs. are moved out; there is a disabled item in the list; embedded button as you've done here; menu stays open after select vs. closes after select, etc).

I'm merging this in, and then I'll add that in another branch. Thanks for this work!

PR #4, opened 2017-12-20
New architecture: the Dispatch type

thomashoneyman:
#### What does this pull request do?

`Dispatch` is a major architectural addition to the project. To describe it, I'll back up a few steps.

1. The purpose of this library is to provide building blocks for creating selection user interfaces.
2. These building blocks are called _primitives_. Our design philosophy expects primitives to all exist as siblings within a parent component that ties them all together. This is done so that you can mix and match primitives instead of using a set of pre-made components without flexibility.
3. One defining feature of primitives is that they do not contain a render function. The parent is free to render the primitive wherever they want, and to describe how the primitive should render however they want. We do provide some helper functions for this.
4. For this reason, we have a clear separation between the _events_ that trigger state transitions in a primitive, and those actual transitions. The `Container` has no idea how it was toggled, for example; all it knows is how to toggle on and off if a query comes down.

The render function, however, is the problem. Query types in Halogen are an algebraic data type with a necessary polymorphic variable stuck to the end (`a`):

```purescript
data Query a
  = Something a
  | SomethingElse a
```

Every component has a query type, and because these are simple ADTs, they can't be merged together. If you want to take `ModuleA.Query` and stick it to `ModuleB.Query`, you have to wrap them: `ModuleA.Query (ModuleB.Query)`.

The `render` function for a component has the component's state type as its input and the component's query type as its output. However, since the user is writing these render functions, that means _they'll_ be forced to do all this wrapping and pollute their own query algebra with various ways of wrapping child queries within each other in various orders. It also would force them to provide the slots in the render functions so primitives could send queries to other primitives.

This is an enormous mess, but it is solvable by _giving all primitives the same query type_. This is what `Dispatch` does:

```purescript
-- Here's the idea: Create a single `Dispatch` type that serves as the query type for ALL
-- primitives. In any given primitive, case on your specific query type, and return all
-- others wrapped in Emit.

data Dispatch item o a
  = ParentQuery (o Unit) a
  | S (SearchQuery Unit) a
  | C (ContainerQuery item Unit) a

------
-- Primitives

-- The Search primitive
data SearchQuery a
  = TextInput String

-- The Container primitive
data ContainerQuery item a
  = Highlight   Target              -- Change the highlighted item
  | Select      Int                 -- Select a particular item
  | Key         KeyboardEvent       -- A key has been pressed
  | Mouse       MouseState          -- Update mousedown state
  | Blur                            -- Blur event
  | Visibility  VisibilityStatus    -- Open or close the menu
  | SetItems    (Array (Item item)) -- Set the data (used by parent)
```

Now we can freely mix various types together in the same render function:

```purescript
getInputProps = augmentHTML
  [ HE.onFocus      $ HE.input_ $ C $ Visibility Toggle
  , HE.onKeyDown    $ HE.input  $ \ev -> C $ Key ev
  , HE.onValueInput $ HE.input  $ \ev -> S $ TextInput ev
  , HE.onMouseDown  $ HE.input_ $ C $ Mouse Down
  , HE.onMouseUp    $ HE.input_ $ C $ Mouse Up
  , HE.onBlur       $ HE.input_ $ C $ Blur
  , HP.tabIndex 0
  ]
```

and we can allow the parent to embed whatever queries _they_ want with a simple helper:

```purescript
embed :: ∀ item parentQuery. H.Action parentQuery -> Unit -> Dispatch item parentQuery Unit
embed = ParentQuery <<< H.action

-- in use; fortunately this gives a type error if the valueInput event is not supported:
getInputProps [ HE.onMouseOut $ HE.input_ $ embed MyQuery ]
```

This is the final key to making primitives work effectively, but it's still quite rough.

#### Where should the reviewer start?

* Read the new examples readme so you can easily run the examples: [readme](https://github.com/citizennet/purescript-halogen-select/blob/916ba91e4330d5ee3fcd63e17612a6fd817e311c/examples/readme.md)
* Read through the new Dispatch file: [dispatch](https://github.com/citizennet/purescript-halogen-select/blob/916ba91e4330d5ee3fcd63e17612a6fd817e311c/src/Dispatch.purs)
* Look at the Dropdown diff to see how this changes what the end user has to do in practice: [Container -> Dropdown](https://github.com/citizennet/purescript-halogen-select/compare/dispatch?expand=1#diff-cf473f108b57958aaf0e939f737e9d31)
* Look at the diffs for the `Container` and `Search` primitives: [Container](https://github.com/citizennet/purescript-halogen-select/compare/dispatch?expand=1#diff-3defdd4802cefd6f6ffda198ff6b5f4e), [Search](https://github.com/citizennet/purescript-halogen-select/blob/916ba91e4330d5ee3fcd63e17612a6fd817e311c/src/Primitive/Search.purs)

#### How should this be manually tested?

1. Build all the examples from the root with `npm run all-examples`.
2. Open each example and verify the dropdown works and the typeahead filters by search.

At this point, these implementations are minimal and `master` is a record of the current state of the repo. Later, we'll have more involved testing here because the `master` branch will deploy a new package on Pursuit.

#### Other Notes:

**Documentation**
The project documentation will need to be updated to describe how the `Dispatch` type works. With this and the external render functions, we have everything we need to build an initial version.

**Quality of Life Improvements**
There are a number of places where we should find ways to improve our code. The major stumbling block is how all primitive query types have to be handled by all primitives. I have opened an issue about it.


thomashoneyman:
This is the issue referenced above in "quality of life": handling the duplicate code when a primitive emits other primitives' messages back to the parent:

https://github.com/citizennet/purescript-halogen-select/issues/5

thomashoneyman:
Latest commit adds debounce functionality to the typeahead.
https://github.com/citizennet/purescript-halogen-select/blob/dispatch/src/Primitive/Search.purs#L58

davezuch (approved):

PR #6, opened 2018-01-10
Replace component individual states with Store comonad & add calendar example

thomashoneyman:
### What does this pull request do?

This pull request has two major changes:

1. It adds a new architecture for primitive component types, relying on the `Store` comonad to allow us to pass updated render functions to the components;
2. It adds a new day picker example in the `calendar/` folder

**Render functions have been moved to the `Input` type for a primitive.**
The most significant change in the project is that render functions are no longer passed as arguments to components when they are mounted in a slot; instead, the render functions are part of any primitive's `Input` type. This is necessary because of a restriction in Halogen:

> After a component is initialized in a slot, providing a different value for any parameter other than the input value or slot address will have no effect on the child component. [Halogen Guide](https://github.com/slamdata/purescript-halogen/blob/master/docs/5%20-%20Parent%20and%20child%20components.md#rendering)

In other words, once a slot has been initialized, we have no way to update the way it renders. This matters when you're passing in values that need to change depending on the *parent's* state. For example, in a day picker, the primitive renders the grid of items, but it'll also render the month and year the user selected. Those are carried on the parent's state, not the primitive. When the parent re-renders, we need to update the render function with the new values, but we can't -- unless we pass the render function in the child's `Input` type.

**Render functions cannot be carried in `State`.**
The primitive's state type would result in cycles if we put the render function in the type synonym. Instead, with a suggestion from @monoidmusician (Nicholas Scheel), I've used a glorified tuple called the `Store` comonad to store the component's state and render function separately. The implementation can be seen in the [Dispatch file](https://github.com/citizennet/purescript-halogen-select/compare/calendar?expand=1#diff-5a9ec4678a785e78138bc1a98a7166c1), [Container primitive](https://github.com/citizennet/purescript-halogen-select/compare/calendar?expand=1#diff-3defdd4802cefd6f6ffda198ff6b5f4e), and [Search primitive](https://github.com/citizennet/purescript-halogen-select/compare/calendar?expand=1#diff-b741b62e351ba3605f4a71d6d75743ce).

**With render as part of input, we have to move state & input types to `Dispatch`**
This can be seen in the Dispatch diff.

### Where should the reviewer start?

1. Read through the changes to `Dispatch`, `Container`, and `Search`
2. Build the `calendar` example and test out its behavior.

### Other Notes:

This is a major change to the project, and with this change, our documentation is out of date. Before the library settles on a v0.1, I want to fully update the documentation with a tutorial on creating your own example component using this new architecture as it may not be immediately obvious to everyone. There is also the change where render functions are passed directly in `Input`.


src/Dispatch.purs
```diff
@@ -51,23 +89,40 @@ data VisibilityStatus
...
+
+
+emit :: ∀ a0 a1 o item e f. Applicative f => (o Unit -> f Unit) -> Dispatch item o e a0 -> a1 -> f a1
+emit f (ParentQuery o _) a = a <$ f o
+emit _ _ a = pure a
+
+updateState :: ∀ state html. (state -> state) -> Store state html -> Store state html
+updateState f inputStore =
+  let (Tuple r oldState) = runStore inputStore
+   in store r $ f oldState
```
MonoidMusician: Is this just `seeks`? https://pursuit.purescript.org/packages/purescript-transformers/3.5.0/docs/Control.Comonad.Store#v:seeks

thomashoneyman: You're right! Updated in latest commit. Thanks for the note!

src/Dispatch.purs
```diff
@@ -18,24 +27,53 @@ to share query types.
 -- primitives. In any given primitive, case on your specific query type, and return all
 -- others wrapped in Emit.
 
-data Dispatch item o a
+data Dispatch item o e a
```
thomashoneyman: We're forced to carry this `e` around everywhere because it's the effect type necessary for the `SearchState` and thus the `SearchInput` render function. This is because the search state carries an effect around with it due to the implementation of the debouncer. It's a bit annoying because it's a heavy cost to carry another type parameter every single place `Dispatch` is used.

However, I found [this code from @garyb](https://github.com/slamdata/purescript-halogen/issues/479#issuecomment-331702420) that may work; this would obviate the need to carry this parameter everywhere.

MonoidMusician: Yeah, I was thinking about this ... I think the other option would be passing `Dispatch item o e a` around along with `item` (for `Container`) ... I guess `SearchState` would still require `e` though, but in general it looks like each thing doesn't need access to all four parameters, but mostly just a way to create a `Dispatch item o e a`.

Or give up and use IO :trollface:

src/Primitive/Container.purs
```diff
@@ -19,125 +23,127 @@ The Container primitive ...
...
-  { items            :: Array (Item item)
-  , open             :: Boolean
-  , highlightedIndex :: Maybe Int
-  , lastIndex        :: Int
-  , mouseDown        :: Boolean
-  }
-
-type Input item =
-  { items :: Array (Item item) }
+type State item o e = Store (ContainerState item) (H.ComponentHTML (Dispatch item o e))
```
thomashoneyman: It should be possible to move this into Dispatch, where the `State` for any primitive takes that primitive's state type and constructs the `Store`.

src/Primitive/Container.purs
```diff
@@ -19,125 +23,127 @@ The Container primitive ...
...
+             H.modify $ seeks (_ { open = true })
+          Off    -> do
+             H.liftAff $ log "set off"
+             H.modify $ seeks (_ { open = false, highlightedIndex = Nothing })
+          Toggle ->  do
+             H.liftAff $ log "toggle"
+             H.modify $ seeks (\st -> st { open = not st.open, highlightedIndex = Nothing })
+
+        ContainerReceiver i -> a <$ do
+          -- Replaces the state entirely with a newly-initialized one.
```
thomashoneyman: Remove; inaccurate.

src/Primitive/Container.purs
```diff
@@ -19,125 +23,127 @@ The Container primitive ...
...
+        H.raise $ Emit (Search q unit)
 
-      C q a -> case q of
+      Container q a -> case q of
         Select index -> do
-          st <- H.get
-          if not st.open then pure a else a <$ case st.items !! index of
-            Just (Selectable item) -> H.raise $ ItemSelected item
-            _ -> H.liftAff $ log $ "Cannot select item at that index: " <> show index
+          (Tuple _ st) <- pure <<< runStore =<< H.get
```
thomashoneyman: Consider replacing with a function in Dispatch called `getState` (or similar) that replaces `pure <<< runStore =<< H.get`

src/Primitive/Container.purs
```diff
@@ -19,125 +23,127 @@ The Container primitive ...
...
-            eval $ C (Visibility Off) unit
+            eval $ Container (Visibility Off) unit
 
         -- When toggling, the user will lose their highlighted index.
         Visibility status -> a <$ case status of
-          On     -> H.modify (_ { open = true })
-          Off    -> H.modify (_ { open = false, highlightedIndex = Nothing })
-          Toggle -> H.modify \st -> st { open = not st.open, highlightedIndex = Nothing }
+          On     -> do
+             H.liftAff $ log "set on"
```
thomashoneyman: Should remove logging statements for the release of the lib.

src/Primitive/Search.purs
```diff
@@ -8,56 +8,49 @@ import Data.Maybe (Maybe(..), fromMaybe)
...
-      S q a -> case q of
+      Search q a -> case q of
         TextInput str -> do
-          st  <- H.get
-          H.modify _ { search = str }
+          (Tuple _ st) <- pure <<< runStore =<< H.get
+          H.modify $ seeks _ { search = str }
 
           case st.debouncer of
             Nothing -> unit <$ do
```
thomashoneyman: Replace `AVar String` with `AVar Unit` because the internals aren't being used; it solely exists to block.

examples/calendar/src/Calendar.purs
```diff
@@ -0,0 +1,296 @@
...
+component =
+  H.parentComponent
+    { initialState: const initState
+    , render
+    , eval
+    , receiver: const Nothing
+    }
+  where
+    initState :: State
+    initState = { targetDate: Tuple (unsafeMkYear 2018) (unsafeMkMonth 2), statuses: [] }
```
thomashoneyman: Demonstrate getting the current month in this example.

examples/calendar/src/Calendar.purs
```diff
@@ -0,0 +1,296 @@
...
+              [ HP.class_ $ HH.ClassName "flex pv3" ]
+              [ arrowButton (ToggleYear Prev) "<<" (Just "ml2")
+              , arrowButton (ToggleMonth Prev) "<" Nothing
+              , dateHeader
+              , arrowButton (ToggleMonth Next) ">" Nothing
+              , arrowButton (ToggleYear Next) ">>" (Just "mr2")
+              ]
+              where
+                -- Buttons next to the date.
+                arrowButton :: H.Action (Query e) -> String -> Maybe String -> _
```
thomashoneyman: Fill in holes in the type signatures.

src/Primitive/Container.purs
MonoidMusician: `H.gets pos`?

thomashoneyman:
@crcornwell @whoadave Main task list before this will pass your reviews:

1. No library warnings (fill in type signatures)
2. Logging only in examples, not in lib functions

Main task list before release:

1. Document the lib in the readme
2. Provide inline documentation for all core library functions; don't export helpers

PR #16, opened 2018-02-28
SearchContainer focus

davezuch:

This PR fixes focus and blur behavior for the search primitive. It also updates some of the queries to make more sense and exposes more messages to the parent. Specifically, when the container is clicked on (usually because an item has been selected, but the click could come from anywhere inside the container), it sends a message to the parent so that it can re-trigger focus on the search input. Also when the container visibility is set, it sends a message to the parent so that it can do things like trigger validation.

Start with the primitives, then see how those changes affect the docs components.

## Other Notes:

Project documentation need to be updated. Specifically the markdown tutorials are out of sync with how the docs components are implemented.

### DIFF
--- a/docs/src/Components/Dropdown/Dropdown.purs
+++ b/docs/src/Components/Dropdown/Dropdown.purs
@@ -80,6 +80,8 @@ component =
       HandleContainer m a -> case m of
         C.Emit q -> eval q *> pure a
 
+        C.ContainerClicked -> pure a
+
         C.ItemSelected item -> do
           H.liftAff $ log ("Selected: " <> item)
           H.modify \st -> st { selected = ( item : st.selected ) }
--- a/docs/src/Components/Dropdown/Dropdown.purs
+++ b/docs/src/Components/Dropdown/Dropdown.purs
@@ -5,7 +5,7 @@ import Prelude
 import Control.Monad.Aff.Class (class MonadAff)
 import Control.Monad.Aff.Console (log, logShow, CONSOLE)
 import DOM (DOM)
-import Data.Array ((:), difference, mapWithIndex, length, delete)
+import Data.Array (mapWithIndex)
 import Data.Maybe (Maybe(..))
 import Halogen as H
 import Halogen.HTML as HH
@@ -17,7 +17,7 @@ type DropdownItem = String
 
 type State =
   { items    :: Array DropdownItem
-  , selected :: Array DropdownItem }
+  , selected :: Maybe DropdownItem }
 
 data Query a
   = Log String a
@@ -39,7 +39,7 @@ component =
     }
   where
     initState :: State
-    initState = { items: testData, selected: [] }
+    initState = { items: testData, selected: Nothing }
 
     render
       :: State
@@ -80,33 +80,28 @@ component =
       HandleContainer m a -> case m of
         C.Emit q -> eval q *> pure a
 
-        C.ContainerClicked -> pure a
-
         C.ItemSelected item -> do
           H.liftAff $ log ("Selected: " <> item)
-          H.modify \st -> st { selected = ( item : st.selected ) }
+          H.modify \st -> st { selected = pure item }
 
           st <- H.get
           H.liftAff $ logShow st.selected
-          _  <- H.query unit
-                  $ H.action
-                  $ C.ContainerReceiver
-                  $ { render: renderContainer
-                    , items: difference st.items st.selected }
+          _ <- H.query unit
+                 $ H.action
+                 $ C.SetVisibility C.Off
           pure a
+
+        otherwise -> pure a
 
       Removed item a -> do
         st <- H.get
-        let newSelections = delete item st.selected
-            newItems = difference newSelections st.items
-
-        H.modify (_ { selected = newSelections })
+        H.modify (_ { selected = Nothing })
 
         _  <- H.query unit
                 $ H.action
                 $ C.ContainerReceiver
                 $ { render: renderContainer
-                  , items: newItems }
+                  , items: st.items }
 
         pure a
 
@@ -118,6 +113,7 @@ HELPERS
 
 -}
 
+class_ :: ∀ p i. String -> H.IProp ( "class" :: String | i ) p
 class_ = HP.class_ <<< HH.ClassName
 
 -- The parent must provide some input data.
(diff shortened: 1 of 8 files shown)


docs/src/Components/Dropdown/Dropdown.purs
```diff
@@ -82,29 +82,26 @@ component =
...
         st <- H.get
-        let newSelections = delete item st.selected
-            newItems = difference newSelections st.items
-
-        H.modify (_ { selected = newSelections })
+        H.modify (_ { selected = Nothing })
 
         _  <- H.query unit
                 $ H.action
                 $ C.ContainerReceiver
```
thomashoneyman: Now that there's a `ReplaceItems` query, this ought to be replaced with that.

`H.query unit $ H.action $ C.ReplaceItems st.items`

src/Primitives/Container.purs
```diff
@@ -110,13 +111,18 @@ type ContainerInput o item =
 -- | - `ItemSelected`: An item has been selected in the container. This does not indicate the item
 -- |                   has been removed; if you would like the item to also be removed from the
 -- |                   container, make sure to query `ReplaceItems` from the parent.
+-- | - `ContainerClicked`: The container has been clicked, which draws focus away from the
+-- |                       search if used with a search primitive, this message lets the search
+-- |                       container know to return focus to it
```
thomashoneyman, line 117: Should add the doc for `VisibilitySet` also.

```
- `VisibilitySet`: Notifies the parent that the container visibility has a new value. Useful when you need to trigger some behavior (like validation) when the user closes or opens the menu.
```

src/Primitives/Container.purs
```diff
@@ -210,20 +216,30 @@ component =
...
+            Up   -> do
+              H.modify $ seeks (_ { mouseDown = false })
+              H.raise ContainerClicked
 
       Blur a -> do
         (Tuple _ st) <- getState
         if not st.open || st.mouseDown
           then pure a
-          else a <$ (eval $ Visibility Off a)
+          else a <$ (eval $ SetVisibility Off a)
```
thomashoneyman: I like this pattern of triggering eval queries rather than duplicate logic, even if simple. 👍

src/Primitives/Container.purs
```diff
@@ -210,20 +216,30 @@ component =
...
 
       -- When toggling, the user will lose their highlighted index.
-      Visibility status a -> a <$ case status of
-        On     -> H.modify $ seeks (_ { open = true })
-        Off    -> H.modify $ seeks (_ { open = false, highlightedIndex = Nothing })
-        Toggle -> H.modify $ seeks (\st -> st { open = not st.open
-                                              , highlightedIndex = Nothing })
+      SetVisibility status a -> a <$ do
+        case status of
+          On     -> do
```
thomashoneyman: Any reason why not `H.modify $ seeks (_ { open = true })`? @whoadave

thomashoneyman: `do` block is redundant

src/Primitives/Container.purs
```diff
@@ -336,28 +352,22 @@ getContainerProps :: ∀ o item e.
      (H.IProp
         ( onMouseDown :: ET.MouseEvent
         , onMouseUp :: ET.MouseEvent
-        , onBlur :: ET.FocusEvent
```
thomashoneyman, line 339: 👍 Glad to see these events being removed. Less is better as far as listeners go.

src/Primitives/Search.purs
```diff
@@ -198,6 +219,19 @@ component =
       -- the parent to set the text.
       SearchReceiver i a -> H.modify (updateStore i.render id) *> pure a
 
+      CaptureFocus e a -> a <$ do
+        (Tuple _ st) <- getState
+        let el = either (const Nothing) Just <<< runExcept <<< readHTMLElement <<< toForeign <<< currentTarget <<< focusEventToEvent $ e
```
thomashoneyman: This is pretty long. What about:

```purescript
let el = either (const Nothing) Just 
        <<< runExcept 
        <<< readHTMLElement 
        <<< toForeign 
        <<< currentTarget 
        <<< focusEventToEvent 

H.modify $ seeks _ { inputEl = el e }
H.raise Focused
```

src/Primitives/Search.purs
```diff
@@ -198,6 +219,19 @@ component =
...
+        H.modify $ seeks _ { inputEl = el }
+        H.raise Focused
+
+      TriggerFocus a -> do
+        (Tuple _ st) <- getState
+        case st.inputEl of
+          Nothing -> pure a
+          Just el -> do
+            _ <- H.liftEff $ focus el
+            pure a
```
thomashoneyman: We run into effectful cases like this often @crcornwell @whoadave @foresttoney. In these cases we can make things a bit cleaner with traverse:
```purescript
TriggerFocus a -> a <$ do
  (Tuple _ st) <- getState 
  traverse_ (H.liftEff <<< focus) st.inputEl
```

src/Primitives/Search.purs
```diff
@@ -228,20 +260,16 @@ getInputProps :: ∀ o item e eff
...
         | e
         )
         (SearchQuery o item eff)
       )
 getInputProps = flip (<>)
-  [ HE.onFocus      $ HE.input_ $ FromContainer $ H.action $ C.Visibility C.Toggle
+  [ HE.onFocus      $ HE.input  $ \ev -> CaptureFocus ev
   , HE.onKeyDown    $ HE.input  $ \ev -> FromContainer $ H.action $ C.Key ev
   , HE.onValueInput $ HE.input  TextInput
-  , HE.onMouseDown  $ HE.input_ $ FromContainer $ H.action $ C.Mouse C.Down
```
thomashoneyman: 👍

thomashoneyman (approved):
@whoadave I have a few suggestions here for some minor improvements, but all of them are stylistic. Functionally this is good to go. If you want, I can go ahead and merge and make any of these updates and release a new patch version.

thomashoneyman:
Note: the CI build is failing because `select` relies on `ocelot` which relies on `select`...

Once we update the typeahead component over there the build will work again.

src/Primitives/Search.purs
thomashoneyman: Also -- should we check this? Ensure that there isn't already a reference stored? Or would we always want to overwrite just in case?

src/Primitives/Search.purs
crcornwell: This could also be:
```
[ HE.onFocus      $ HE.input CaptureFocus
, HE.onKeyDown    $ HE.input $ FromContainer <<< H.action <<< C.Key
```

src/Primitives/Search.purs
crcornwell: hush!

src/Primitives/Container.purs
```diff
@@ -385,30 +395,27 @@ getItemProps :: ∀ o item e
...
-        , tabIndex :: Int
         | e
         )
         (ContainerQuery o item)
       )
 getItemProps index = flip (<>)
-  [ HE.onClick     $ HE.input_ $ Select index
+  [ HE.onMouseDown $ HE.input_ $ Select index
   , HE.onMouseOver $ HE.input_ $ Highlight (Index index)
   , HE.onKeyDown   $ HE.input  $ \ev -> Key ev
```
crcornwell:
```
HE.onKeyDown $ HE.input Key
```

src/Primitives/Container.purs
```diff
@@ -277,7 +293,7 @@ getToggleProps :: ∀ o item e f.
           f
        )
 getToggleProps q = flip (<>)
-  [ HE.onClick      $ HE.input_ $ q $ H.action $ Visibility Toggle
+  [ HE.onClick      $ HE.input_ $ q $ H.action $ ToggleVisibility
   , HE.onKeyDown    $ HE.input  $ \ev -> q $ H.action $ Key ev
```
crcornwell, line 296:
```
HE.onKeyDown $ HE.input $ q <<< H.action <<< Key
```

crcornwell (approved):

PR #18, opened 2018-03-05
Replace `Search` and `Container` primitives with `Select` component

thomashoneyman:
This pull request introduces a major overhaul to the library's architecture based on our experience using it to build typeaheads and dropdowns in `purescript-ocelot`. The breaking point came when we needed to always throw focus back to the input element from the container after an action is taken -- this means so much bi-directional information passing that there is no use having separate components for the two.

In addition, it turns out we'll want to capture and debounce input streams even when using just a toggle input -- for things like highlighting `California` when a user types `Cal-`, like browsers do.

This refactor has led to a much simpler API and use while preserving the ability to render as you see fit. It also fixes some focus / blur bugs and includes minor behavior improvements (like highlighting the first entry when you search so you can hit `Enter` right away for the best match).

In my opinion, this is a necessary and core change to the way the library works. It turns out we'll never have a container in isolation -- it will always have a driver, either a toggle or a text input, and the vast majority of their behaviors will be the same. 

The new structure:

1. A `Select` component that can be input-driven (search your items) or toggle-driven (capture key events, but no searching)
2. A `Utilities.Setters` module that exports all the helpers you can use to attach behaviors
3. A `Internal.State` module with helpers for updating `Store`-based components.

1. Review the new `Select` module, which is basically a combined search/container: https://github.com/citizennet/purescript-halogen-select/blob/53e0f9cb54baebf9c420b5ee9044aadf5f9680e4/src/Select.purs
2. Review how this affects the typeahead example: https://github.com/citizennet/purescript-halogen-select/compare/search-or-toggle?expand=1#diff-086954f132a5e0fe4df0e71d3c5d163e

Other examples have been removed for the time being and can be replaced later.

Build the docs and ensure the typeahead works as expected.

## Other Notes:

Before this can be released to Pursuit, the documentation will need an update. This is also necessary in order to integrate the new typeahead updates in `ocelot`.


src/Select.purs
```diff
@@ -0,0 +1,261 @@
...
+        (Tuple _ st) <- getState
+        if st.visibility == Off then pure a else do
+           let prevent = H.liftEff <<< preventDefault <<< KE.keyboardEventToEvent
+           case KE.code ev of
+            "ArrowUp"   -> prevent ev *> (eval $ Highlight Prev a)
+            "ArrowDown" -> prevent ev *> (eval $ Highlight Next a)
+            "Escape"    -> do
+               prevent ev
+               traverse_ (H.liftEff <<< blur) st.inputElement
+               eval $ Blur a
```
davezuch: Test removing this -- should be triggered in the DOM

davezuch:
Before merging, experiment with having the container continuously listen to the parent's items.

PR #22, opened 2018-04-08
Free query

MonoidMusician:

Base the Select's Query off a Free Monad approach. The old `Query` is now called `QueryF`, and `Query` is now `Free (QueryF ...)`. This allows using all of the monadic language to express what should happen when events trigger now.

As an example of putting this to use, I changed the logic in `onMouseDown` in `setToggleProps` to get the current visibility of the element and then choose to blur or focus it based on the current visibility, as well as toggling the visibility. (Now toggling is no longer a primitive!)

I also added `setItemPropsAndBlur` to allow blurring after selecting an item. This behavior was possible before – I think it might be a little more intuitive, but something we can discuss.

There is a little more boilerplate in the `Select` module, to help with the free queries (mostly just adding `liftF` and applying `Unit` by default), but this helps a _tiny_ bit with the usual Halogen boilerplate (`Just <<<` instead of `H.action` I guess). `Select.always` helps with the remaining boilerplate. Note how `toggleVisibility` is now implemented via `getVisibility` and `setVisibility`.

Implementation changes:
- for `Query`: see `toggleVisibility`, `triggerFocus`, and `eval`

Behavior changes:
- As noted above, see `setToggleProps` and `setItemPropsAndBlur`

Otherwise it should be the same code expressed differently ;)

In the documentation site, observe that:
1. Clicking on an item (in either the typeahead or dropdown) will blur the input element.
2. Clicking on the dropdown toggle when it is already open will blur the toggle element.

## Other Notes:

Also added some tips to the readme for building the documentation site. YMMV.


MonoidMusician:
From chat: the typeahead component should probably continue to use `setItemProps`, since it allows multiselect. Might be able to use `raise` and parent communication to allow more custom behavior like this.

thomashoneyman:
Thanks, @MonoidMusician! I've reviewed the code a little myself, and we'll do a more thorough review early this week. I've pinged the other folks at CitizenNet to take a look also. I appreciate the contribution :)

thomashoneyman:
I've spent some time with this and there are parts I love and parts that are too much complexity for the relative gain IMO

**Pros**
- The reduced boilerplate for users in omitting `H.action` and using `Select.doSomething`
- The monadic syntax in the props helpers

**Unsure**
- We end up rewriting all the queries as functions again, which is a bit more boilerplate added to the library. Then again, queries rarely change.
- Any Halogen user is used to the `H.action Query` pattern, and though it's more boilerplate at least it's more familiar than remembering `Select.query` instead. Then again, they'd need to go look up the queries in the first place, so I'm not sure this is really a drawback.

**Cons**
- Using the state monad for visibility seems like overkill; I'd rather stick with your original implementation in your first commit
- Is there a reason to import `purescript-strings`? Or is that unnecessary? I'd like to avoid the extra dependency if possible.

@MonoidMusician with these points in mind, I'm thinking that your first two commits and your last documentation commit make sense to merge (unless you have any additions or improvements you'd like to make!) but I'd prefer to hold off on using the state monad for visibility.

How does that sound?

MonoidMusician:
Okay, thanks for the feedback, that’s totally reasonable! I think I can revise it in a few days.

re `strings`: it was already a transitive dependency, but you started using those NonEmpty strings (released in the latest version) and I’m not sure if bower reflected that. Could totally be a screwup on my end though.

MonoidMusician:
Also, some of my intention with the Free monad pattern is to reveal how Halogen already fits that pattern ;D

In particular this distinction between actions and queries, which Halogen usually treats with the raw constructors (`a -> Query a` and `(a -> a) -> Query a`), already has the commonly accepted convention in free monads of having “setters” return `Unit` (which means applying `unit` for the `a` parameter) and having “getters” return the value being gotten (which ends up with `id` for the `value -> a` part).

And I think this makes sense. So it’s a pattern I would like to encourage, even if it’s a bit unfamiliar to Halogeners. Having a prescribed concrete type to work with is probably easier than trying to figure out “Oh I need `id` here to return the value ...”. Plus, it works great for _Monads_ ;)

thomashoneyman:
Having had more time to explore free monads in the past week I agree with you and I'm happy to adopt the approach here.

thomashoneyman (changes requested):
Marking with changes requested as noted in the discussion so far.

thomashoneyman:
@MonoidMusician Is there a way that I can help out with this? Love to get this merged in soon and I understand completely if you're pressed for time.

MonoidMusician:
Hey I think I was just going to revert the `State` monad commit, clean up a couple things ... the big part that I think you could help with is documenting how to use the new style. E.g. “Instead of `H.action ToggleVisibility` just use `toggleVisibility`” (but of course there are more complex cases too, e.g. combining it with inputs and everything ... haven’t fully thought it through).

Oh and we wanted to revert the blurring at least on the multi-select example.

Saturday I’ll be completely free and should get to it!

MonoidMusician:
just pushed the commits I had sitting here locally …

MonoidMusician:
Okay this is what I got, let me know if it looks good or anything else you want!

thomashoneyman (approved):

PR #27, opened 2018-06-18
Added const version of setter props

davezuch:

Moved all the props in `Setters` to their own const value and then referenced those from the setter functions. That way if I do want to override them, I can do something like:

```
HH.input ( Setters.inputProps <> [ HE.onValueInput $ const Nothing ] )
```

Instead of:

```
HH.input ( Setters.setInputProps [] <> [ HE.onValueInput $ const Nothing ] )
```

It's a very minor ergonomic improvement, but also avoids doing some pointless `flip` and extra `<>` operations.

### DIFF
--- a/src/Utils/Setters.purs
+++ b/src/Utils/Setters.purs
@@ -84,7 +84,12 @@ setInputProps
   :: ∀ o item p
    . Array (HP.IProp (InputProps p) (Query o item Unit))
   -> Array (HP.IProp (InputProps p) (Query o item Unit))
-setInputProps = flip (<>)
+setInputProps = flip (<>) inputProps
+
+inputProps
+  :: ∀ o item p
+   . Array (HP.IProp (InputProps p) (Query o item Unit))
+inputProps =
   [ HE.onFocus \ev -> Just do
       Select.captureRef $ FE.toEvent ev
       Select.setVisibility On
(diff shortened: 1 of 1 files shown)


thomashoneyman:
I see how this makes it easier to override properties when you want to ensure you're changing the underlying behavior. On the other hand, I'm hesitant to introduce `props / setProps` for every pair when the difficulty introduced is simply to save the single empty array operation:

```purescript
HH.input ( Setters.inputProps <> [ HE.onValueInput $ const Nothing ] )
```

saves a little effort in the empty array, but introduces two functions / values you ought to know about instead of just one

```purescript
HH.input ( Setters.setInputProps [] <> [ HE.onValueInput $ const Nothing ] )
```

and only works specifically when you have no properties of your own you want to add and purely want to override. for example, in the below case, you're back to using `setInputProps`:

```purescript
HH.input ( Setters.setInputProps [ css "my-class" ] <> [ HE.onValueInput $ const Nothing ] )
```

unless you don't care about overwriting properties -- which you should care about! Ideally properties are rarely overwritten, and if you really want a different set, you simply replace the array altogether. I understand that introduces a bit of code duplication and your suggestion here makes it nicer to do those partial overrides.

I think remaining with just `setInputProps` is more clear, and if you want to override you can provide the empty array and then concatenate (or provide an array of properties and then concatenate).

**tl;dr version**
I'm wary of adding more functions / values if it's very easy and clear how to achieve the same result with what already exists in the library. Are you sure that this update introduces enough benefit for the extra cognitive overhead?

PR #31, opened 2018-11-01
Fix a typo

alexeyzab:
I noticed a typo while going through the tutorial and decided to fix it.
Without this change you would get the following error:

```
Error found:
in module Component
at src/Component.purs line 66, column 59 - line 66, column 61

  Unknown value st

```


davezuch (approved):
Nice catch!

PR #33, opened 2018-11-08
Input/Toggle Reference

crcornwell:

This change improves the way that `Select` focuses or blurs its input or toggle element. Previously we would capture a reference to the element by relying on the focus event and storing that reference on state. This change makes `setInputProps` and `setToggleProps` set a `ref` on the element, which can be queried for in order to focus or blur.

## Other Notes:

As the commits suggest, I went through a few different iterations before arriving here. Knowing that these `RefLabel`s don't have to be unique, I think this is the cleanest approach. The `setToggleProps` and `setInputProps` signatures don't need to change, we can get rid of the `CaptureRef` query and relying on focus to get a reference to the element, and we can even get rid of the `inputElement` reference on the state. Plus it fits nicely into our existing expectation of the consumer, which is that they'll have to use those various `setProps` helpers to augment their properties; this is just another property that needs to go onto the element.

### DIFF
--- a/src/Select.purs
+++ b/src/Select.purs
@@ -18,17 +18,8 @@ import Effect.Aff (Fiber, delay, error, forkAff, killFiber)
 import Effect.Aff.AVar (AVar)
 import Effect.Aff.AVar as AVar
 import Effect.Aff.Class (class MonadAff)
-import Halogen
-  ( Component
-  , ComponentDSL
-  , ComponentHTML
-  , RefLabel(..)
-  , component
-  , getHTMLElementRef
-  , liftAff
-  , liftEffect ) as H
+import Halogen as H
 import Halogen.HTML as HH
-import Halogen.Query.HalogenM (fork, raise) as H
 import Renderless.State (getState, modifyState_, modifyStore)
 import Web.Event.Event (preventDefault)
 import Web.HTML.HTMLElement (blur, focus)
@@ -260,7 +251,7 @@ component =
       , search: fromMaybe "" i.initialSearch
       , debounceTime: fromMaybe (Milliseconds 0.0) i.debounceTime
       , debouncer: Nothing
-      , inputRef: H.RefLabel "999z"
+      , inputRef: H.RefLabel "select-input"
       , items: i.items
       , highlightedIndex: Nothing
       , visibility: Off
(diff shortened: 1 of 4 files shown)


src/Select.purs
```diff
@@ -353,9 +338,9 @@ component =
           "ArrowUp"   -> preventIt *> eval' (highlight Prev)
           "ArrowDown" -> preventIt *> eval' (highlight Next)
           "Escape"    -> do
-            st <- getState
+            inputElement <- H.getHTMLElementRef $ H.RefLabel "select-input"
```
davezuch, line 341: I wonder if we should move the `RefLabel` to the state just to serve as a const. That way we're not manually typing the label string out whenever we need it, since the compiler isn't going to tell us if we mis-typed it

davezuch: I also wonder if there's much of a performance cost to running `getHTMLElementRef` every time we need the input element

davezuch (approved):
Glad we're making this change, this feels much cleaner

src/Select.purs
crcornwell: Yeah, it'd be nice to have this as a string constant, but having it on the state seems like a heavy price for that since we'd have to have access to the state in those setProps functions, which just felt kinda ugly when I had it in here before.

As for `getHTMLElementRef`, I have to imagine there's some cost to querying the DOM as opposed to just storing the element on state, but it will only be happening on focus, blur, or an `ESC` keypress so I doubt it will be a problem. We can always go back to storing it on state and fetching it during initialization if it does become an issue, but then we may have to think about what happens if that element leaves the DOM and returns at some point.

davezuch: Yeah, it'll be something to look into if there's a noticeable lag for any of those actions, otherwise I don't think it's a big deal.

As for the `RefLabel` on the state, I just meant updating the state to something like:
```purescript
type State item =
  { ...
  , refLabel :: RefLabel
  }

--

initialState i = store i.render
      { ...
      , refLabel : RefLabel "select-input"
      }
```

And then the `Focus` and `Key` evals use that instead of recreating the string. Don't even need it on the state actually, could just make it a const inside the component's `where` block

MonoidMusician: I wouldn't put it in state, but you could just dump it as a string constant in some file: give it a name and always reference it by that name, so the string literal only appears that one place.

PR #35, opened 2018-11-29
Add debouncing via `Ref`

thomashoneyman:

Updates debouncing for the `Select` component to use a reference instead of requiring a state modification every time the debouncer runs. While this doesn't save us anything in the case that the debouncer completes and the action runs, it saves a state modification every time the debouncer is cancelled (the user pressed a key and interrupted the previous timer).

Closes #35 

Review the changes to `Select.purs`, specifically the `Search` query.

I don't have an implementation that's ready to use, so I haven't verified this works as expected. But before merging I'd definitely recommend building Ocelot without this change and then again with the change to verify that it saves the renders as expected. I'm slammed until the weekend, but I can also test myself over the weekend if needed.

### DIFF
--- a/src/Select.purs
+++ b/src/Select.purs
@@ -13,11 +13,13 @@ import Control.Monad.Free (Free, foldFree, liftF)
 import Data.Array (length, (!!))
 import Data.Maybe (Maybe(..), fromMaybe)
 import Data.Time.Duration (Milliseconds(..))
-import Data.Traversable (for_, traverse_)
+import Data.Traversable (for_, traverse, traverse_)
 import Effect.Aff (Fiber, delay, error, forkAff, killFiber)
 import Effect.Aff.AVar (AVar)
 import Effect.Aff.AVar as AVar
 import Effect.Aff.Class (class MonadAff)
+import Effect.Ref (Ref)
+import Effect.Ref as Ref
 import Halogen as H
 import Halogen.HTML as HH
 import Renderless.State (getState, modifyState_, modifyStore)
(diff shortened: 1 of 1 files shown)


src/Select.purs
```diff
@@ -267,27 +271,33 @@ component =
...
+        ref :: Maybe Debouncer <- H.liftEffect $ map join $ traverse Ref.read st.debounceRef
         modifyState_ _ { search = str }
         setVis On
 
-        case st.inputType, st.debouncer of
+        case st.inputType, ref of
           TextInput, Nothing -> unit <$ do
             var   <- H.liftAff AVar.empty
             fiber <- H.liftAff $ forkAff do
               delay st.debounceTime
```
thomashoneyman, line 288: Should probably guard this, too, so we don't even go through any of these steps unless a debounce time has actually been set.

src/Select.purs
```diff
@@ -267,27 +271,33 @@ component =
...
             _ <- H.fork do
               _ <- H.liftAff $ AVar.take var
-              modifyState_ _ { debouncer = Nothing, highlightedIndex = Just 0 }
+              void $ H.liftEffect $ traverse_ (Ref.write Nothing) st.debounceRef
+              modifyState_ _ { highlightedIndex = Just 0 }
               newState <- getState
               H.raise $ Searched newState.search
 
-            modifyState_ _ { debouncer = Just { var, fiber } }
+            void $ H.liftEffect $ traverse_ (Ref.write $ Just { var, fiber }) st.debounceRef
```
thomashoneyman, line 300: This will no longer cause a re-render.

MonoidMusician:
LGTM! Just a tiny note: `void` is redundant when you use `traverse_` :)

PR #37, opened 2019-01-03
Fixed ordering of setters so that user event handlers override defaults

davezuch:

Turns out the last `IProp` for an event handler wins. The reason we thought otherwise is that our lovely `appendIProps` function in Ocelot reverses the order of the props. But anyone not using Ocelot, or at least `appendIProps`, will not be able to override the default props here.

## Other Notes:

I have a [corresponding branch](https://github.com/citizennet/purescript-ocelot/compare/append-iprops-order#diff-5658dc7cefe70576ae75aa9b780dfe6b) in Ocelot that fixes `appendIProps`. Once this PR is approved and a new release tagged, I'll make a PR there as well.

### DIFF
--- a/src/Select/Setters.purs
+++ b/src/Select/Setters.purs
@@ -41,7 +41,7 @@ setToggleProps
   :: ∀ o item p
    . Array (HP.IProp (ToggleProps p) (Query o item Unit))
   -> Array (HP.IProp (ToggleProps p) (Query o item Unit))
-setToggleProps = flip (<>)
+setToggleProps = (<>)
   [ HE.onFocus $ Select.always $ Select.setVisibility On
   , HE.onMouseDown \ev -> Just do
       Select.preventClick ev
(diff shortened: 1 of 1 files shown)


src/Select/Setters.purs
```diff
@@ -118,7 +118,7 @@ setItemProps
    . Int
   -> Array (HP.IProp (ItemProps p) (Query o item Unit))
   -> Array (HP.IProp (ItemProps p) (Query o item Unit))
-setItemProps index = flip (<>)
+setItemProps index = (<>)
```
thomashoneyman, line 121: could just be `append`

thomashoneyman:
I _knew_ something was up!

thomashoneyman (approved):

thomashoneyman:
@linearray I'm assuming this is related to your issue in Slack. Glad @davezuch saw the cause of the problem.

src/Select/Setters.purs
davezuch: True...but actually more verbose lol

linearray:
Thank you for the quick fix!

crcornwell (approved):

Issue #38, opened 2019-03-13
Upgrade to Halogen v5


thomashoneyman:
Select is not compatible with the [release candidate for Halogen 5](https://github.com/slamdata/purescript-halogen/releases/tag/v5.0.0-rc.3) and will prevent users and libraries that depend on Select from upgrading when Halogen is released.

thomashoneyman:
Completed in #39

Issue #36, opened 2018-12-23
<input> inside dropdown doesn't work


rnons:
## Environment

* PureScript **[0.12]**
* Halogen **[master]**
* Select **[5.0 branch]**

## Current behavior

Because `mousedown` is prevented when using `setContainerProps`, text input inside dropdown doesn't work, cannot focus or input text into it.
A reproduce: https://rnons.github.io/halogen-select-issue/

## Expected behavior

Should be able to focus and input text into `<input>` inside dropdown.
In [halogen-day-picker](https://github.com/rnons/purescript-halogen-day-picker/blob/master/src/Halogen/DayPickerInput.purs#L194-L209), I used a `clickedInside` flag to decide visibility.

Since this is a `select` library, it's understandable if you don't want to support this case, just want to make sure. Thanks

davezuch:
Hi @rnons, thanks for opening an issue. So the way Select works, all DOM events that trigger behaviors such as selection highlighting or whether visibility is on or off, they're bound to the [toggle or input props](https://github.com/citizennet/purescript-halogen-select/blob/master/src/Select/Setters.purs#L87-L91). That means those behaviors will only work as long as the toggle or input has focus, which is why we prevent default behavior on `mousedown`, because we don't want focus to leave our toggle/input element as long as the user is clicking inside the container.

If you want to have your search input inside the dropdown, there is a trick we use which you can see in action at the bottom of [this page](https://citizennet.github.io/purescript-ocelot/#typeaheads).

Basically you put your search (with input props attached) inside your dropdown container, but instead of conditionally rendering it, you conditionally show/hide it (we have an `"offscreen"` class that uses absolute positioning to hide an element off-screen). Then you just make your "toggle" element a `label`, which takes advantage of the label's default behavior of triggering focus on the first child input. That means clicking on the `label` will trigger focus on the child search element, which will switch `Visibility` to `On`, and then everything should just work.

A lightweight example:

```purescript
HH.label
  [ css "relative" ]
  [ HH.button_ "toggle"
  , HH.div
      [ css if visibility ==  Off then "offscreen" else "dropdown" ]
      [ HH.input setInputProps myInputProps
      , renderItems
      ]
   ]
```

```css
.relative { position: relative; }
.offscreen { position: absolute; left: -9999%; top: -9999%; }
.dropdown { position: absolute; left: 0; top: 100%; }
```

[...]

PR #39, opened 2019-04-03
Update for Halogen 5

thomashoneyman:

Addresses #38. 

This PR updates Select to be Halogen 5 compatible and makes a number of improvements to the component's extensibility without adding more boilerplate or removing any features. It doesn't change the behavior of the component internally, but the component should now be used differently.

A summary of changes:

1. The component and examples are now Halogen 5 compatible.

2. There is no longer any need to wrap Select in another component to make a selection interface. You can create typeaheads, dropdowns, date pickers, etc. simply by extending the component's state, queries, actions, input, and messages and by hooking into when messages are emitted, the component is initialized or finalized, or when the component receives input.

3. There is, accordingly, no need to provide any parent state or parent queries to the component. Users who were previously doing this can recover the old behavior by extending Select with a parent query or action and a message that re-raises it, instead of or in addition to creating a few more actions meant specifically for Select. There's no loss of functionality.

4. There is no longer a `QueryF` or free query pattern; now that you can easily extend actions without boilerplate there's less reason to build complex chains of binds in HTML, and now that queries are not allowed in HTML there's no way to take the result of a query and pass it to the next. Removing this machinery makes things quite a bit simpler.

5. The component no longer outputs any messages if you don't want it to. You can choose to emit nothing; to emit some new custom messages only; to emit custom messages and Select messages; or to emit only Select messages and no new custom ones. This lets you customize the component and decide what messages are relevant to users of the customized component. Maybe they don't need to know about searches because you've already handled it.

6. There is now a `Spec` type, similar to the `ComponentSpec` type used to make a Halogen component, which comes with a `defaultSpec` implementation. This type lets you provide handlers for the ways you've extended the component. If you added new actions, you'd want to add a `handleAction` function to handle them. You can override `defaultSpec` with only the fields you need; for example, if you've got a simple component with just a new render function, you can do `Select.component $ Select.defaultSpec { render = render }`.

7. While the `Spec` type has a lot of type variables, the individual pieces no longer do. In general, Select types are just a standard type synonym + a type parameter describing how it has been extended. So, for example, you have `Action extraAction`, `Query extraQuery childSlots`, `State extraState`, `Input extraState`, etc. Even better, if you are implementing a new `Spec`, you can just implement all your extensions local to the body of that function, so you don't even have to specify the types of anything except for the `Spec` itself and the `Query` type (to tell the compiler where the `a` comes from).

8. There is no longer any need for the Store comonad and the `StateStore` type, nor any need for Renderless as a library. The comonad was needed so that you could pass the render function via `Input` but not keep it in state and not introduce any wrapping types or wrapping HTML. You had to pass the render function via `Input` because the way to extend Select's state was by embedding parent state. And if parent state is embedded, then you need for parent state changes to cause state changes in Select, and that only happens via `Input`. Now that you can extend Select's state freely, you don't need to embed parent state and you don't need to pass the render function via Input and you don't need `StateStore` or Renderless.

9. There is now a `default.nix` file that allows anyone with the Nix package manager to have all dependencies (including mkdocs) to work on the repository by running `nix-shell`.

<details>
  <summary>Previous summary of changes</summary>

This will be a little stream of consciousness as I have just a few minutes, but more specifically:

- Updates the component and examples to be Halogen 5 compatible
- Replaces the existing 'free query' implementation with an extensible variant for actions (queries are no longer permitted in HTML, and actions are not functors, so no free monad available; this _could_ be worked around by wrapping all actions in a query). This also removes the `Raise` and `Emit` pattern; you can simply embed your actions from the parent and trigger them from the render function. See the typeahead for an example.
- Replaces the existing messages with an extensible variant; this enables users to now trigger and handle their own custom messages from the component as they see fit.
- In keeping with Halogen 5, Select now supports embedding further child components into a selection interface. This makes it possible to put, say, a text component inside a dropdown (helps with issues like #36). No example here, but see the [Renderless updates](https://github.com/thomashoneyman/purescript-halogen-renderless/pull/2) for how this is done. You can also send queries through Select to the child component like you can in Formless.
- Preserves only the queries that return responses. The query algebra isn't extensible, so if you want to add an action and then call it from a parent you can do that with the `PerformAction` query.
- Introduces a `Simple` module which provides type synonyms for the core Select types for when you don't need to extend with any additional messages, actions, or child components. Not sure if this is worth keeping around or not.

This PR also updates the documentation and examples as far as the code goes; I haven't updated documentation describing the library and how it works.

## Main Issues

1. The biggest change here is the switch to extensible variants for messages and actions. As I worked on Renderless it seemed significantly cleaner to allow a user to extend the actions and messages with additional ones to their heart's delight -- and provide a handler for the new actions -- than to require them to add new behaviors to a parent component and then wrap them in a `Raise` query and handle them in their own `eval` when the `Emit` message showed up. This keeps the extensibility contained in `Select` without leaking out to the parent. I think this is definitely the right move to make. 

2. One decision I didn't make was to make `State` extensible. This is possible, too -- it would allow us to do things like make a date picker that maintains extra information without that info leaking out and having to be managed by the parent. Or imagine a typeahead that manages the items for itself but is just a simple extension to Select. It might be worth looking into, but I didn't explore it yet.

3. I don't know if it is possible to make the query algebra extensible. Maybe with some funky `VariantF` manipulation, but I don't know how to make the `a` type parameters unify. Maybe it would be better to make the query algebra extensible, instead of the action type, and then provide an action type that can evaluate a `Query Unit` so they can still be used in HTML. That would let us get our free query pattern back. But I don't know how to do this atm. Maybe @MonoidMusician has ideas?

4. The documentation is going to need a serious update to accommodate these changes. I'm going back into a busy period, so I won't be able to put a ton of time into this for a while.

5. Select is still deficient in a few ways -- never fixed the accessibility issues, you can move the highlight beyond the edge of the container and it doesn't scroll you down, etc. -- and I'd like to at least address some of them as part of a new breaking release. Might as well not have to make _two_ breaking releases. If you have any ideas for what has been deficient in CitizenNet usage so far (some queries never used? some clunky? could have used some new ones? not enough information in state?)

</details>


bower.json
```diff
@@ -31,8 +31,10 @@
     "generated-docs"
   ],
   "dependencies": {
-    "purescript-halogen": "^4.0.0",
-    "purescript-halogen-renderless": "^0.0.3"
+    "purescript-halogen": "^5.0.0-rc.3",
+    "purescript-halogen-renderless": "^0.0.4",
+    "purescript-variant": "^6.0.0",
+    "purescript-fixed-points": "^5.1.0"
```
thomashoneyman: @davezuch this was for my brief foray into removing the newtype from `Action`, inspired by https://discourse.purescript.org/t/exploring-mutually-recursive-variants/424. if there's no sense going that direction (i'm feeling like there's not) then i'll remove the dependency.

examples/Components/Typeahead/Typeahead.purs
```diff
@@ -2,178 +2,198 @@ module Docs.Components.Typeahead where
...
-
 type TypeaheadItem = String
 
-data Query a
-  = Log String a
-  | HandleInputContainer (Select.Message Query TypeaheadItem) a
-  | Removed TypeaheadItem a
+data Action
+  = Remove TypeaheadItem
+  | HandleSelect (Select.Message TypeaheadItem ())
```
thomashoneyman: This is an example of how the variant is much nicer than the `Raise` / `Emit` pattern. This component no longer has to handle the new action (log) as part of its actions and handler. Instead, Select now handles this action for itself, which makes more sens.

examples/Components/Typeahead/Typeahead.purs
```diff
@@ -2,178 +2,198 @@ module Docs.Components.Typeahead where
...
+      , items: difference st.items st.selected
+      , render: renderSelect
+      }
+
+  handleExtraActions 
+    :: Variant ExtraActions 
+    -> H.HalogenM _ (Select.Action TypeaheadItem ExtraActions () m) () _ m Unit
+  handleExtraActions = match  
+    { log: \str -> Console.log str 
+    }
```
thomashoneyman: Here's where the new action handler is written (and then passed to Select on line 73). This is what Select will use to handle the new `log` action.

examples/Components/Typeahead/Typeahead.purs
```diff
@@ -2,178 +2,198 @@ module Docs.Components.Typeahead where
...
+        if state.visibility == Select.Off
+          then []
+          else [ renderItems $ renderItem `mapWithIndex` state.items ]
       where
-        renderChild =
-          HH.div
-          [ HE.onClick $ Select.always $ Select.raise $ H.action $ Log "I was clicked" ]
+      renderChild =
+        HH.div
+          [ HE.onClick \_ -> Just $ log "I was clicked" ]
```
thomashoneyman: The newly-extended action can be used just like any other Select action now. (You can also recursively call other Select actions in the handler).

examples/Components/Typeahead/Typeahead.purs
```diff
@@ -2,178 +2,198 @@ module Docs.Components.Typeahead where
...
-      Removed item a -> do
-        st <- H.get
-        H.modify_ _ { selected = filter ((/=) item) st.selected }
-        newSt <- H.get
-        let newItems = difference newSt.items newSt.selected
-        _ <- H.query unit $ Select.replaceItems newItems
-        pure a
+    Remove item -> do
+      newSt <- H.modify \oldSt -> oldSt { selected = filter ((/=) item) oldSt.selected }
+      void $ H.query _select unit $ H.tell $ Select.ReplaceItems $ difference newSt.items newSt.selected
```
thomashoneyman: If we make `State` extensible, then this action could just be an extension of `Select` and the required state (the selected items list) could also just be an extension. Then there's no wrapper component required at all -- you have a typeahead that's still just the `Select` component.

src/Select.purs
```diff
@@ -22,124 +25,94 @@ import Effect.Ref (Ref)
...
 import Web.UIEvent.MouseEvent as ME
 
-----------
--- Component Types
+-----
+-- ACTIONS
 
--- | A useful shorthand for the Halogen component type
-type Component o item m = H.Component HH.HTML (Query o item) (Input o item) (Message o item) m
+newtype Action item v ps m = Action (Variant
```
thomashoneyman: This newtype is necessary because of the self-reference in `andThen`. There is also a mutual reference with `Input` because of `receive` which is a record type synonym, so if this isn't newtyped then at least that will need to be.

src/Select.purs
```diff
@@ -22,124 +25,94 @@ import Effect.Ref (Ref)
...
--- | Sets the component with new input.
-receive :: ∀ o item . Input o item -> Query o item Unit
-receive i = liftF (Receive i unit)
-
--- | Initializes the component on mount.
-initialize :: ∀ o item. Query o item Unit
-initialize = liftF (Initialize unit)
+  -- send a query through `Select` to a child component
+  | SendQuery (ChildQueryBox ps (Maybe a))
+  | PerformAction (Action item v ps m) a
```
thomashoneyman: `PerformAction` makes all actions accessible from the parent via `H.query`. Not sure if this is desirable; perhaps some subset of actions could be split out and made accessible while others are private (receive, initialize, preventClick, key, andThen come to mind as ones that don't have much use being called from a parent).

examples/Components/Dropdown/Dropdown.purs
```diff
@@ -2,100 +2,104 @@ module Docs.Components.Dropdown where
...
-        _ <- H.query unit $ Select.replaceItems $ difference st.items [ item ]
-        H.modify_ _ { text = item }
-        pure a
+  handleAction :: Action -> H.HalogenM State Action (ChildSlots m) Message m Unit
+  handleAction = case _ of
+    HandleSelect msg -> onMatch
+      { selected: \item -> do
+          st <- H.get
+          _ <- H.query _select unit $ H.tell $ Select.PerformAction $ Select.setVisibility Select.Off
+          _ <- H.query _select unit $ H.tell $ Select.ReplaceItems $ difference st.items [ item ]
```
davezuch: Hmm, if we update our helpers then less of this has to change
```purescript
setVisibility = H.tell <<< PerformAction <<< SetVisibility
```
```purescript
_ <- H.query _select unit $ Select.setVisibility Select.Off
```

thomashoneyman: But that would preclude the use of `setVisibility` in the render function, which is of type `H.ComponentHTML Action ChildSlots m`. You can't use queries in components anymore.

An alternative is to make everything a query, and then have an action `PerformQuery (Query Unit)` that lets you use the action-style ones. But then we get back into the extensibility problem.

thomashoneyman:
Major updates in the last couple of commits:

1. I ended up implementing extensible state after all. It has a huge benefit: if state is extensible, then you don't need to embed values from parent state into the render function. If you don't need to embed parent state into Select then you don't need to pass the render function via the `Input` type (it doesn't need to be updated on parent re-render). If you don't need to pass the render function via `Input`, then you can dramatically simplify the type signatures involved and you can remove the `Store` comonad (and the `Renderless` dependency altogether), even while making the component much more extensible than it was before. Doing this incurs two new requirements: any extended fields in State have to be passed via Input (the compiler will check this for you); for any fields that are present in State but not in Input and therefore are going to be inserted (for example, `lastIndex` is created in the initialState function, but is not part of the Input record) we incur a `Row.Lacks` constraint to ensure the extended fields do not contain any of the fields we're going to insert. Both of these are pretty common sense requirements, however. 

2. I removed `Variant` altogether for simplicity's sake. Now, the Message, Query, Action, etc. types are standard ADT's just as you would see in ordinary Halogen. This means that you have to do a little more wrapping and unwrapping, but because there is still no `Raise` / `Emit` pattern, this is much less tedious than it was before. If you take a look at the examples you can see that the wrapping is occasionally annoying, but compared to the mental effort required by people fairly new to PureScript in understanding how variants work, I thought it was worth it. I do like the elegance of Variant, but I'm just not sure the benefit is strong enough in this case.

3. I did discover how to make `Query` extensible after all, and so I've switched from an extensible `Action` type to an extensible `Query` type. I thought this was quite a bit nicer because it allows the `Action` type to remain internal and because queries are more powerful. Now that you can extend the query algebra you can do both action-style extensions and query-style extensions that return information about the component (like `GetCurrentSelections` in a typeahead). In my opinion this is much better than just extending actions. Since you can't use queries in HTML anymore, there's a `RunQuery` constructor that can be used to wrap action-style queries for use in HTML.

4. The last thing preventing these components from being extensible without making a wrapper component was to figure out odd cases like when an item is selected or when a search is performed. You can't just extend the query algebra and trigger some custom handler when a search action happens, for example, because the whole point there is that the search is debounced before any further work is done. Messages are currently the solution; they exist to signify some important event has occurred and let something else run a handler based on that event (the parent component). So I've changed Select to run a custom handler in addition to raising any message that it sends out. That way you can 'subscribe' to these messages from within the component itself; in a typeahead, you can now listen for a search to filter the list of visible items, without introducing a wrapping component. That function is called `handleMessage` and is another pretty simple one.

5. I simplified the examples. Take a look at the dropdown and typeahead and notice how you can get a working version of both components without ever introducing another component apart from Select itself. 

-----

Extensibility could go even farther but I might be willing to draw the line here. For example, in some components you might want to be able to hook in to the initializer and run something at that point, or whatever. That's the biggest drawback right now: you cannot hook in and modify any existing logic in the component, like handlers for particular actions, unless you replace the action with your own, copy the code out, and modify it yourself. Then again, this is _always_ the case in functions, which are opaque, so it's not really an unexpected issue.

Curious to hear what y'all think of the changes. Maybe someone could try adapting the Ocelot typeahead to this branch of Select and see how it goes?

src/Select.purs
```diff
@@ -288,18 +289,15 @@ handleAction handleExtraQuery handleMessage = case _ of
...
+    handleAction handleExtraQuery handleMessage act2 
     pure unit
   
   RunQuery query -> do
     _ <- handleQuery handleExtraQuery query
     pure unit
   
   where
-  -- recursively evaluate an action
-  handleAction' = handleAction handleExtraQuery handleMessage
```
thomashoneyman: This is interesting: defining this helper function causes a "too much recursion" error when the code is run. But not defining this helper and instead writing it out every time fixes the issue. I'd like to try to find a smaller reproduction as it's an irritating restriction.

thomashoneyman: Ah -- no, as the next commit shows, you can also eta-expand it to get around the issue. Still annoying.

examples/Components/Dropdown/Dropdown.purs
```diff
@@ -3,99 +3,128 @@ module Docs.Components.Dropdown where
...
-  , text :: String }
+{-
+  There is no longer any need to write a new component to wrap Select. Instead, you can
+  extend Select's state, input, query algebra (and by extension, actions), messages, and 
+  child components. Rather than create a wrapping component, simply use Select with  
+  your extensions in place!
+
+  This module exports a version of Select extended to be a proper dropdown. It renders
+  a button and maintains the current selection from the list.
+-}
```
davezuch: So, why the two different comment styles (`{- ... -}` here, `-- ...` below)?

src/Select.purs
```diff
@@ -195,189 +133,223 @@ data InputType
...
+  initialState :: Input item st -> State item st
+  initialState input = Builder.build pipeline input
+    where
+    pipeline =
+      Builder.modify (SProxy :: _ "debounceTime") (fromMaybe (Milliseconds 0.0))
+        >>> Builder.modify (SProxy :: _ "search") (fromMaybe "")
+        >>> Builder.insert (SProxy :: _ "debounceRef") Nothing
+        >>> Builder.insert (SProxy :: _ "visibility") Off
+        >>> Builder.insert (SProxy :: _ "highlightedIndex") Nothing
+        >>> Builder.insert (SProxy :: _ "lastIndex") (length input.items - 1)
```
davezuch: I like this way of handling optional inputs 👍

src/Select.purs
```diff
@@ -195,189 +133,223 @@ data InputType
...
+      TextInput, Nothing -> unit <$ do
+        var   <- H.liftAff AVar.empty
+        fiber <- H.liftAff $ forkAff do
+          delay st.debounceTime
+          AVar.put unit var
+
+        -- This compututation will fork and run in the background. When the
+        -- var is finally filled, the action will run (raise a new search)
+        _ <- H.fork do
+          _ <- H.liftAff $ AVar.take var
```
davezuch: Why not `void $` everywhere instead of switching between styles here? Or does that not work here?

examples/Components/Typeahead/Typeahead.purs
```diff
@@ -2,178 +2,247 @@ module Docs.Components.Typeahead where
...
+      }
+  
+  Select.Searched str -> do
+    st <- H.get
+    let 
+      visible = difference st.selectedItems $ filter (contains (Pattern str)) st.items
+      index = elemIndex str visible
+    H.modify_ _ { visibleItems = visible }
+    for_ index \ix -> do
+      Select.handleAction handleQuery handleMessage $ Select.Highlight $ Select.Index ix
```
davezuch: Feels a bit clunky having to pass these along. Though I guess if users find it annoying they can create their own helper
```purescript
handleAction = Select.handleAction handleQuery handleMessage
```

examples/Components/Typeahead/Typeahead.purs
```diff
@@ -2,178 +2,247 @@ module Docs.Components.Typeahead where
...
+        , closeButton item
+        ]
 
     closeButton item =
       HH.span
-      [ HE.onClick $ HE.input_ (Removed item)
-      , class_ "absolute pin-t pin-b pin-r p-1 mx-3 cursor-pointer" ]
+        [ HE.onClick \_ -> 
+	    -- nicer with variants. worth it?
+            Just $ Select.RunQuery $ Select.Embed $ Remove item unit
```
davezuch: A little 😉

davezuch (approved):
> I removed Variant altogether for simplicity's sake. Now, the Message, Query, Action, etc. types are standard ADT's just as you would see in ordinary Halogen. This means that you have to do a little more wrapping and unwrapping, but because there is still no Raise / Emit pattern, this is much less tedious than it was before. If you take a look at the examples you can see that the wrapping is occasionally annoying, but compared to the mental effort required by people fairly new to PureScript in understanding how variants work, I thought it was worth it. I do like the elegance of Variant, but I'm just not sure the benefit is strong enough in this case.

I didn't feel like the use of variants made things that much more complicated. On the one hand, this is already a pretty advanced library, so most users trying to come in and tinker will probably be familiar with variants. I do see your argument for not making it even more complicated, I'm just not sure you're gaining much simplicity with a standard ADT, as the complexity just moves to dealing with the embedded queries and messages. I found the variant approach rather elegant. I'm not asking you to revert back though, just sharing my thoughts on the matter.

>Extensibility could go even farther but I might be willing to draw the line here. For example, in some components you might want to be able to hook in to the initializer and run something at that point, or whatever. 

Hmm, if we were still extending actions instead of queries, we should be able to just have the initializer be a parameter that's passed in. That way if they want a custom initializer, they can extend the actions with it and pass it in, otherwise they'd just pass in `Select.initialize`.

Overall, this is super nice! I would've been happy if it was a simple upgrade to Halogen 5, but you went and made tremendous improvements! Looking forward to seeing just how much nicer our other codebases get as a result. Thanks for putting in all this effort @thomashoneyman!

src/Select.purs
thomashoneyman: I use `void` more nowadays but I haven't gone through to fix the rest of the code here. I'll switch all of it to one style or the other before merging. Have a preference?

examples/Components/Typeahead/Typeahead.purs
thomashoneyman: Can't do that ^ -- unfortunately it leads to infinite recursion if you don't eta-expand. i don't know why. i'm probably going to file a ticket.

thomashoneyman:
@davezuch @crcornwell @foresttoney I am prepared to merge these changes early next week (Tuesday?) into Select's master branch in preparation for the Halogen 5 release. Let me know if you see anything worrying in the meantime.

There are more changes necessary before Select will be ready for a new release. Specifically:

1. We have a lot of documentation to update to this new style. As part of that effort, I would like to add a little new documentation explaining how to recover the old behavior of Select (extending parent queries and emitting them as messages) for folks who were relying on that behavior before (@rnons, for example). Otherwise, we should be able to trim down the overall docs.

2. I haven't updated the actual module documentation yet, which will also need to happen. I haven't updated it because I'm planning another set of changes to Select's internals that will make it more usable in more scenarios (like component-in-component, which is currently useless because we maintain visibility based on focused state). Those will be coming in a separate PR which can be reviewed on its own.

Please consider messing around with the new styled examples by building the documentation site and going to the examples page. You should see that they still work as before, except better: the typeahead is asynchronous and the dropdown actually filters out selected items. If you have Nix installed you can use these commands to get that running right away:

```sh
nix-shell
yarn && yarn watch-docs
```

Looking forward to any feedback y'all have, and if there is none, then I'll merge this and keep going on improvements in another branch.

davezuch (approved):
I already approved these updates in the PR on your fork, but I'll say it again here, really nice work @thomashoneyman!

> I'm planning another set of changes to Select's internals that will make it more usable in more scenarios (like component-in-component, which is currently useless because we maintain visibility based on focused state). Those will be coming in a separate PR which can be reviewed on its own.

I'm interested to see these changes. Any progress on them yet?

thomashoneyman:
> I'm interested to see these changes. Any progress on them yet?

Not yet, I haven't had time. I want to update Formless before going any further on Select so that I can update Real World; once those updates are all ready I'd like to return to Select.

PR #43, opened 2019-04-22
Pass state to `setContainerProps` call

i-am-the-slime:
I was just trying this and got this as a compile error. Passing state fixes it so I suppose that's correct.

### DIFF
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -66,7 +66,7 @@ spec = S.defaultSpec { render = render, handleMessage = handleMessage }
 
     renderContainer = whenElem (st.visibility == S.On) \_ ->
       HH.div
-        ( SS.setContainerProps [ class_ "Dropdown__container" ] )
+        ( SS.setContainerProps st [ class_ "Dropdown__container" ] )
         ( renderItem `mapWithIndex` st.items )
       where
       renderItem index item =


thomashoneyman:
Hey @i-am-the-slime -- thanks for taking the time to post an issue. I worked on the setters file recently and I don't believe the old version (for Halogen 4) or the new version (for Halogen 5) require passing state to any of the setters. For reference:

Halogen 5 (master)
https://github.com/citizennet/purescript-halogen-select/blob/21d63d1239c8859e15440103a4442a52b52242ff/src/Select/Setters.purs#L122

Halogen 4
https://github.com/citizennet/purescript-halogen-select/blob/c646a5502a7f5e7790e6d1977feb34216c08f9d3/src/Select/Setters.purs#L132

I'm not sure what would cause this error, but perhaps you simply need to delete your `output` folder and rebuild in case there are some stale modules in there? If that doesn't work, do you mind letting me know?

thomashoneyman:
Closing this as `setContainerProps` should not require state.

i-am-the-slime:
Thanks for having a look at this @thomashoneyman. Maybe there was indeed a caching problem, however this is a new project that never depended on an old version of Halogen or Halogen Select. I will have time to try this again over the weekend.

i-am-the-slime:
I am sorry, I made a mistake. The actual place that `st` was needed.
I see you have changed this though:
https://github.com/citizennet/purescript-halogen-select/commit/21d63d1239c8859e15440103a4442a52b52242ff
Looks like everything's fine!

thomashoneyman:
Yes, that was a bad slip on my part. Glad to hear things are working and thanks for reporting this!

PR #45, opened 2019-05-19
Expose `input` type in Select components and combine `spec` and `input` into `Select.component`

### LINKED ISSUE #44: Follow Halogen? Use full type names in type signature
Does this library's Halogen 5 version (e.g. `v5.0.0-rc.2`) intend to follow the spirit of slamdata/purescript-halogen#562?

As a bad example, `act` could become `action`. `act` by itself is already pretty clear, but a few more letters don't hurt, right?
As a better example, `ps` could become `slots`. Unless you learned Halogen 4, `ps` is unclear until you get a compiler error that indicates what goes there or you look at the source code.

Slightly related, but perhaps changing `st` to `stateRows`?

JordanMartinez:
Fixes #44 (not the main focus of this PR, but was added later)

Before this PR, adding a select component as a child in a parent component would look like this:
```purescript
HH.slot label childIndex (S.component Component.spec) (Component.input input) messageHandler
```
This differs from the way Halogen would normally add a child in that only a component would be specified:
```purescript
HH.slot label childIndex component input messageHandler
```
It seems these two things (`Component.spec` and `Component.input`) were separated because one cannot provide a "default value" in `Select.defaultSpec` for the `initialState :: input -> Select.State stateRows` function. If one did add such a function to `Select.defaultSpec`, it ruins the record-updating syntax.
- Without that function, we'd write `S.defaultSpec { render = ... }`
- With that function, we'd write `(S.defaultSpec initialstate) { render = ... }`

This pull request makes two changes.

First, `Select.Component` has an additional argument `(input -> Input st)`. Now one can write
```purescript
selectComponent :: H.Component HH.HTML S.Query' Input Message Aff
selectComponent = S.component initialState $ S.defaultSpec
  { render = -- ...
  }
  where
    initialState :: input -> Input st
    initialState = -- implementation

    -- etc.
```
which looks a lot more like the Halogen component code we're used to. Moreover, it enables authors of this library to provide a "Select Component Template file" (similar to [`ComponentTemplate.purs`](https://github.com/JordanMartinez/learn-halogen/blob/latestRelease/src/05-ComponentTemplate-B.purs)) that people can copy-and-paste to help them get started with using this library.

Second, `Select.Spec` and `Select.Spec'` now take an additional type argument--`input`--and its `receive` function now receives a value of type `input` rather than `Input st`

Third, the Dropdown and Typeahead examples have been updated to use this new component-like syntax rather than the separate-spec-and-input syntax. 

Finally, my editor automatically removes extra whitespace in the files. Rather than fight it, I decided to add these as separate commits to make other commits cleaner to read.

Look at the examples folder's `Main.purs` and `Dropdown.purs`/`Typeahead.purs` files to see the difference. Then look at the `Select.component` function.

Since this exposes what was hidden, I don't foresee a need to add any tests.

## Other Notes:

This PR could be further modified to include a "Select Component Template" file. **Edit:** Such a file was later added in this PR.

### DIFF
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -32,6 +32,9 @@ type Input =
   , buttonLabel :: String
   }
 
+component :: H.Component HH.HTML S.Query' Input Message Aff
+component = S.component input spec
+
 input :: Input -> S.Input State
 input { items, buttonLabel } =
   { inputType: S.Toggle
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -33,52 +33,54 @@ type Input =
   }
 
 component :: H.Component HH.HTML S.Query' Input Message Aff
-component = S.component input spec
-
-input :: Input -> S.Input State
-input { items, buttonLabel } =
-  { inputType: S.Toggle
-  , search: Nothing
-  , debounceTime: Nothing
-  , getItemCount: length <<< _.items
-  , items
-  , buttonLabel
-  , selection: Nothing
-  }
-
-spec :: S.Spec State (Const Void) Void () Message Aff
-spec = S.defaultSpec { render = render, handleMessage = handleMessage }
+component = S.component input $ S.defaultSpec
+    { render = render
+    , handleMessage = handleMessage
+    }
   where
-  handleMessage = case _ of
-    S.Selected ix -> do
-      st <- H.get
-      let selection = st.items !! ix
-      H.modify_ _ { selection = selection, visibility = S.Off }
-      H.raise $ SelectionChanged st.selection selection
-    _ -> pure unit
+    input :: Input -> S.Input State
+    input { items, buttonLabel } =
+      { inputType: S.Toggle
+      , search: Nothing
+      , debounceTime: Nothing
+      , getItemCount: length <<< _.items
+      , items
+      , buttonLabel
+      , selection: Nothing
+      }
 
-  render st =
-    HH.div
-      [ class_ "Dropdown" ]
-      [ renderToggle, renderContainer ]
-    where
-    renderToggle =
-      HH.button
-        ( SS.setToggleProps [ class_ "Dropdown__toggle" ] )
-        [ HH.text (fromMaybe st.buttonLabel st.selection) ]
+    handleMessage :: S.Message -> H.HalogenM (S.State State) S.Action' Message Aff Unit
+    handleMessage = case _ of
+      S.Selected ix -> do
+        st <- H.get
+        let selection = st.items !! ix
+        H.modify_ _ { selection = selection, visibility = S.Off }
+        H.raise $ SelectionChanged st.selection selection
+      _ -> pure unit
 
-    renderContainer = whenElem (st.visibility == S.On) \_ ->
+    render :: S.State State -> H.ComponentHTML S.Action' () Aff
+    render st =
       HH.div
-        ( SS.setContainerProps [ class_ "Dropdown__container" ] )
-        ( renderItem `mapWithIndex` st.items )
+        [ class_ "Dropdown" ]
+        [ renderToggle, renderContainer ]
       where
-      renderItem index item =
+      renderToggle =
+        HH.button
+          ( SS.setToggleProps [ class_ "Dropdown__toggle" ] )
+          [ HH.text (fromMaybe st.buttonLabel st.selection) ]
+
+      renderContainer = whenElem (st.visibility == S.On) \_ ->
         HH.div
-          ( SS.setItemProps index
-              [ classes_
-                  [ "Dropdown__item"
-                  , "Dropdown__item--highlighted" # guard (st.highlightedIndex == Just index)
-                  ]
-              ]
-          )
-          [ HH.text item ]
+          ( SS.setContainerProps [ class_ "Dropdown__container" ] )
+          ( renderItem `mapWithIndex` st.items )
+        where
+        renderItem index item =
+          HH.div
+            ( SS.setItemProps index
+                [ classes_
+                    [ "Dropdown__item"
+                    , "Dropdown__item--highlighted" # guard (st.highlightedIndex == Just index)
+                    ]
+                ]
+            )
+            [ HH.text item ]
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -49,7 +49,7 @@ component = S.component input $ S.defaultSpec
       , selection: Nothing
       }
 
-    handleMessage :: S.Message -> H.HalogenM (S.State State) S.Action' Message Aff Unit
+    handleMessage :: S.Message -> H.HalogenM (S.State State) S.Action' () Message Aff Unit
     handleMessage = case _ of
       S.Selected ix -> do
         st <- H.get
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -2,7 +2,6 @@ module Components.Dropdown where
 
 import Prelude
 
-import Data.Const (Const)
 import Effect.Aff (Aff)
 import Data.Array ((!!), mapWithIndex, length)
 import Data.Maybe (Maybe(..), fromMaybe)
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -33,53 +33,53 @@ type Input =
 
 component :: H.Component HH.HTML S.Query' Input Message Aff
 component = S.component input $ S.defaultSpec
-    { render = render
-    , handleMessage = handleMessage
-    }
+  { render = render
+  , handleMessage = handleMessage
+  }
   where
-    input :: Input -> S.Input State
-    input { items, buttonLabel } =
-      { inputType: S.Toggle
-      , search: Nothing
-      , debounceTime: Nothing
-      , getItemCount: length <<< _.items
-      , items
-      , buttonLabel
-      , selection: Nothing
-      }
+  input :: Input -> S.Input State
+  input { items, buttonLabel } =
+    { inputType: S.Toggle
+    , search: Nothing
+    , debounceTime: Nothing
+    , getItemCount: length <<< _.items
+    , items
+    , buttonLabel
+    , selection: Nothing
+    }
 
-    handleMessage :: S.Message -> H.HalogenM (S.State State) S.Action' () Message Aff Unit
-    handleMessage = case _ of
-      S.Selected ix -> do
-        st <- H.get
-        let selection = st.items !! ix
-        H.modify_ _ { selection = selection, visibility = S.Off }
-        H.raise $ SelectionChanged st.selection selection
-      _ -> pure unit
+  handleMessage :: S.Message -> H.HalogenM (S.State State) S.Action' () Message Aff Unit
+  handleMessage = case _ of
+    S.Selected ix -> do
+      st <- H.get
+      let selection = st.items !! ix
+      H.modify_ _ { selection = selection, visibility = S.Off }
+      H.raise $ SelectionChanged st.selection selection
+    _ -> pure unit
 
-    render :: S.State State -> H.ComponentHTML S.Action' () Aff
-    render st =
+  render :: S.State State -> H.ComponentHTML S.Action' () Aff
+  render st =
+    HH.div
+      [ class_ "Dropdown" ]
+      [ renderToggle, renderContainer ]
+    where
+    renderToggle =
+      HH.button
+        ( SS.setToggleProps [ class_ "Dropdown__toggle" ] )
+        [ HH.text (fromMaybe st.buttonLabel st.selection) ]
+
+    renderContainer = whenElem (st.visibility == S.On) \_ ->
       HH.div
-        [ class_ "Dropdown" ]
-        [ renderToggle, renderContainer ]
+        ( SS.setContainerProps [ class_ "Dropdown__container" ] )
+        ( renderItem `mapWithIndex` st.items )
       where
-      renderToggle =
-        HH.button
-          ( SS.setToggleProps [ class_ "Dropdown__toggle" ] )
-          [ HH.text (fromMaybe st.buttonLabel st.selection) ]
-
-      renderContainer = whenElem (st.visibility == S.On) \_ ->
+      renderItem index item =
         HH.div
-          ( SS.setContainerProps [ class_ "Dropdown__container" ] )
-          ( renderItem `mapWithIndex` st.items )
-        where
-        renderItem index item =
-          HH.div
-            ( SS.setItemProps index
-                [ classes_
-                    [ "Dropdown__item"
-                    , "Dropdown__item--highlighted" # guard (st.highlightedIndex == Just index)
-                    ]
-                ]
-            )
-            [ HH.text item ]
+          ( SS.setItemProps index
+              [ classes_
+                  [ "Dropdown__item"
+                  , "Dropdown__item--highlighted" # guard (st.highlightedIndex == Just index)
+                  ]
+              ]
+          )
+          [ HH.text item ]
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -18,7 +18,7 @@ type Slot =
 
 type State =
   ( items :: Array String
-  , selection :: Maybe String 
+  , selection :: Maybe String
   , buttonLabel :: String
   )
 
@@ -32,7 +32,7 @@ type Input =
   , buttonLabel :: String
   }
 
-input :: Input -> S.Input State 
+input :: Input -> S.Input State
 input { items, buttonLabel } =
   { inputType: S.Toggle
   , search: Nothing
@@ -54,8 +54,8 @@ spec = S.defaultSpec { render = render, handleMessage = handleMessage }
       H.raise $ SelectionChanged st.selection selection
     _ -> pure unit
 
-  render st = 
-    HH.div 
+  render st =
+    HH.div
       [ class_ "Dropdown" ]
       [ renderToggle, renderContainer ]
     where
@@ -70,13 +70,12 @@ spec = S.defaultSpec { render = render, handleMessage = handleMessage }
         ( renderItem `mapWithIndex` st.items )
       where
       renderItem index item =
-        HH.div 
-          ( SS.setItemProps index 
-              [ classes_ 
+        HH.div
+          ( SS.setItemProps index
+              [ classes_
                   [ "Dropdown__item"
                   , "Dropdown__item--highlighted" # guard (st.highlightedIndex == Just index)
                   ]
               ]
-          ) 
+          )
           [ HH.text item ]
-
--- a/examples/Components/Dropdown.purs
+++ b/examples/Components/Dropdown.purs
@@ -34,7 +34,7 @@ type Input =
 component :: H.Component HH.HTML S.Query' Input Message Aff
 component = S.component input $ S.defaultSpec
   { render = render
-  , handleMessage = handleMessage
+  , handleEvent = handleEvent
   }
   where
   input :: Input -> S.Input State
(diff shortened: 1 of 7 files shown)


thomashoneyman:
Hey @JordanMartinez -- thanks for opening this PR. I agree that this is a more natural way to use components when you're used to the usual Halogen style. I haven't had time to review thoroughly because I'm currently on vacation, but I look forward to doing so soon.

One thing that's nice about keeping the `Spec` and `Input` types separate (as separate arguments) is that you can then modify the spec downstream if you would like. For example, you might want to write a dropdown component with a default render function and a custom `handleAction`, but leave it open to be overriden by later uses:

```purescript
dropdownSpec = S.defaultSpec 
  { render = defaultDropdownRender
  , handleAction = handleAction 
  }

differentHandleAction = case _ of ...
newRenderFunction state = HH.div ...

-- Same dropdown, but with different rendering
newDropdownSpec = dropdownSpec { render = newRenderFunction }

-- Same dropdown, but with a different handler for actions
otherDropdownSpec = dropdownSpec { handleAction = differentHandleAction }
```

Then again, maybe this is just too much customization (how often will people realistically want to override a component's `handleAction`?).

I'll have to consider the initialState changes / additional `input` type separately. As a rule I'd like to keep type parameters to a minimum.

Once again thanks for taking the time to work on this! I'd appreciate hearing from the folks at CitizenNet (@davezuch, etc.) as well and I'll be sure to get back to you by early next week.

JordanMartinez:
> One thing that's nice about keeping the Spec and Input types separate (as separate arguments) is that you can then modify the spec downstream if you would like. For example, you might want to write a dropdown component with a default render function and a custom handleAction, but leave it open to be overriden by later uses:

Mm... I see your point. On one hand, it might be too much customization. On another hand, `Select.component` (as it's written in this PR) still allows one to do that:
```purescript
-- normal
HH.slot _child unit selectComp input msgHandler

-- customized according to both args
HH.slot _child unit (S.component initialState spec) input msgHandler

-- customized according to spec arg
HH.slot _child unit (comp spec) input msgHandler
  where
    comp spec = S.component initialState spec

-- customized according to initialState arg
HH.slot _child unit (comp initialState) input msgHandler
  where
    comp initialState = S.component initialState spec
```
The `where` clauses above could be defined in a separate file

examples/Components/Dropdown.purs
```diff
@@ -32,51 +31,55 @@ type Input =
...
+    input :: Input -> S.Input State
+    input { items, buttonLabel } =
+      { inputType: S.Toggle
+      , search: Nothing
+      , debounceTime: Nothing
+      , getItemCount: length <<< _.items
+      , items
+      , buttonLabel
+      , selection: Nothing
+      }
```
thomashoneyman: Minor style nit: while it may not be perfectly applied across this library, I prefer not to indent in `where` blocks and to only indent two spaces at a time; I can also make this adjustment everywhere I notice it after the PR lands in a followup commit.

For example:

```purescript
component = S.component input $ S.defaultSpec
  { render = render
  , ...
  }
  where
  input =
    { inputType: S.Toggle
    , ...
    }

File: src/Select.purs [hunk 2]
```diff
@@ -115,47 +115,48 @@ type Input st =
...
       :: Maybe (Action act)
   }
 
-type Spec' st m = Spec st (Const Void) Void () Void m
+type Spec' st input m = Spec st (Const Void) Void () input Void m
 
-defaultSpec :: forall st query act ps msg m. Spec st query act ps msg m
-defaultSpec = 
+defaultSpec :: forall st query act ps input msg m
+             . Spec st query act ps input msg m
```
thomashoneyman: Formatting here is a bit funky; I'd prefer:

```purescript
defaultSpec :: forall st query act ps input msg m. Spec st query act ps input msg m
defaultSpec = ...

-- or if that's too long
defaultSpec
  :: forall st query act ps input msg m
   . Spec st query act ps input msg m
defaultSpec = ...
```

thomashoneyman:
I'd like to rephrase your changes in my own words to make sure we're on the same page. In this PR you introduce:

1. A new type variable, `input`, which represents a user-provided type that will serve as the component's input type to downstream users (for example: if Select is used to make a dropdown, this type will be what users of the _dropdown_ provide as component input).
2. A new argument to the component, a function of type `(input -> Input st)`, which maps the user-provided `input` type to a valid Select `Input` type to produce the initial state.

The reason to introduce the type variable and the extra function is to make the component input customizable. For instance, if you extend Select to support the state and behaviors of a dropdown component, then with this change you could also export a modified input type that reflects what a dropdown component would need, not necessarily what Select needs -- perhaps just a list of items, with sensible defaults set for the rest of Select's usual input fields.

The previous code did not require the type parameter or initial state function. Instead, if you wanted a different input type for a component you'd have to export it along with a helper function from `input -> Input st` that could be applied to the input type before it was passed to the component. For example:

```purescript
type NewInput = ...

mkInput :: NewInput -> Select.Input ExtraState
mkInput = ...

-- old
HH.slot ... component (mkInput input) ...

-- new
HH.slot ... (component mkInput) input ...

-- or with partial application
HH.slot ... component' input ...
  where
  component' = component mkInput
```

On the whole I think this change is reasonable. I was concerned about the `receiver` function only using the custom `input` type and not the full `Input st` type, but if your component only accepts `input` as input then the rest of the values can't change anyway so there's no sense in the receiver depending on them. You could always read the values from state.

I have one thought worth considering before merging, however: 

#### The initial state function is a separate argument. 

Given that we already have this `Spec` with a record of functions in it, and the `Spec` type already has to take the type parameter for `input`, and the spec is already near-identical to the Halogen `ComponentSpec` which takes an initial state function, I'd like to get close to that component spec here. Specifically, I would like for the `initialState` function to be a part of the spec record instead of being a separate argument and for it to have a default value.

The reason you haven't done this in this PR is because there's no way to produce a value of `Input st` by default when `st` is not the empty row. If `st` _is_ the empty row then we ought to be able to produce the regular Select input type, and if `st` is not the empty row then the user is simply going to have to provide an `initialState` function.

But I do wonder whether we could still place this function in the spec record and make it work when `st` is the empty row. You'd get a type error if you try to provide extra state but don't provide an `initialState` function. Something similar occurs with `handleMessage`; if you set the message type to `Void` then you don't have to provide the function, but if you set it to something else you get a type error.

Before merging this I would like to explore whether it is possible to move the initial state function into the spec record.

src/Select.purs
JordanMartinez: I'll go with the second one.

examples/Components/Dropdown.purs
JordanMartinez: That's fine. I've accounted for that in latest commits.

JordanMartinez:
> I'd like to rephrase your changes in my own words to make sure we're on the same page.

Your rephrasing is correct.

> But I do wonder whether we could still place this function in the spec record and make it work when st is the empty row. You'd get a type error if you try to provide extra state but don't provide an initialState function. Something similar occurs with handleMessage; if you set the message type to Void then you don't have to provide the function, but if you set it to something else you get a type error.

There's a few ideas I had when thinking about this. Since none of them were ideal, I went with the "make it an argument to `Select.component` idea."

While we can probably figure out what the end-user's additional state rows will be via type-level programming, we don't know what each label's corresponding type's "default value" will be.

First, I thought we could force the user's extended state types to always be wrapped in a `Maybe`. `Maybe` would make all of their values a `Monoid`. So, via type-level programming to determine what the records labels are, we could create the initial state record via `Nothing`/`mempty` values for the user-added labels. However, that's not ideal because it forces an unnecessary constraint on the end-user. Why pay for the additional overhead of `Maybe` when it's never needed?

Second, I thought we could have a "normal" function and a "helper" function (similar to `Spec` and `Spec'`): 
1. `Select.component`: a function that exposes the `initialState` function should one want to extend the select component with additional state. In other words what we currently have right now.
2. `Select.component'`: a function whose type signature forces `st` to be the empty row and which calls `Select.component` with `const defaultSelectState` as its `initialState` function. `defaultSelectState` would have the type, `Input ()`.

While this is possible, it still isn't ideal because it doesn't make the component "template-able." Halogen 4 used to have multiple functions for creating a component: `component`, `lifecycleComponent`, etc. They decided to get rid of those multiple functions and switched to a single component-creating function that is configured via `Spec`s. Since they switched to that approach, it also made it easy to provide a template file (like the `ComponentTemplate.purs` file linked above) that one can copy, paste, configure, and run without dealing with compiler errors.

I wasn't able to think of other ideas for solving this. Of those that I did, I don't think they are as simple as just adding the `initialState` function to the `Select.component` function.

JordanMartinez:
Latest commit adds an example for what the "template file" for a toggle-based select component might look like.

JordanMartinez:
Latest commit now fixes #44. I'll update the PR's opening comment to include this as well.

src/Select.purs
```diff
@@ -115,47 +115,49 @@ type Input st =
...
     -- perform some action when the component initializes.
-  , initialize 
-      :: Maybe (Action act)
+  , initialize
+      :: Maybe (Action action)
 
     -- optionally perform some action on initialization. disabled by default.
   , finalize
-      :: Maybe (Action act)
+      :: Maybe (Action action)
```
thomashoneyman: If we change this record to be this instead...

```purescript
type Spec st query action slots input msg m =
  { -- usual Halogen component spec
    render :: State st -> H.ComponentHTML (Action action) slots m

    -- handle additional actions provided to the component
  , handleAction :: action -> H.HalogenM (State st) (Action action) slots msg m Unit

    -- handle additional queries provided to the component
  , handleQuery :: forall a. query a -> H.HalogenM (State st) (Action action) slots msg m (Maybe a)

    -- handle messages emitted by the component; provide H.raise to simply
    -- raise the Select messages to the parent.
  , handleMessage :: Message -> H.HalogenM (State st) (Action action) slots msg m Unit

    -- optionally handle input on parent re-renders
  , receive :: input -> Maybe action

    -- perform some action when the component initializes.
  , initialize :: Maybe action

    -- optionally perform some action on initialization. disabled by default.
  , finalize :: Maybe action
  }
```

...then we can change the `Initialize` data constructor to be

```purescript
data Action 
  = ...
  | Initialize (Maybe action)
```

and the `handleAction` for initialization to be...

```purescript
  Initialize -> do
    ...
    for_ mbAction (Action >>> handle)
```

and the component spec to be:

```purescript
component mkInput spec = H.mkComponent
  { initialState: initialState <<< mkInput
  , render: spec.render
  , eval: H.mkEval $ H.defaultEval
      { handleQuery = handleQuery spec.handleQuery
      , handleAction = handleAction spec.handleAction spec.handleMessage
      , initialize = Just (Initialize spec.initialize)
      , receive = map Action <<< spec.receive
      , finalize = map Action spec.finalize
      }
  }
  where
  initialState :: Input st -> State st
  initialState = Builder.build pipeline
    ...
```

You had noticed that these functions map to `Action action`, which means a Select action, instead of just to `action`, which means a user-provided action type. This change would fix that.

Also, I would prefer to name the `input -> Input st` something other than `initialState` and keep the `Input st -> State st` function named `initialState`. In this example I've renamed the new `input -> Input st` function simply `mkInput`, but you may have a better name for it.

thomashoneyman:
> First, I thought we could force the user's extended state types to always be wrapped in a Maybe. 
> ...
> Second, I thought we could have a "normal" function and a "helper" function (similar to Spec and Spec')
> ...
> While this is possible, it still isn't ideal.

Yea, I don't think these are good options either. 

Passing the function seems like the right way to go after all, especially because it will usually be partially-applied. That's something which will need to feature prominently in the new docs. Plus, because this change makes custom input types more convenient, this partial application is worth its price: it can often be invisible to the end consumer and the convenient `Input` more than makes up for it.

Well, I'm going to leave this for a day or two to let anyone on the CitizenNet team weigh in but otherwise I think this is good to go. 

One last thing: do you mind removing the component template file for the time being and opening that as a separate PR? I'd like to consider it separate to this change. Thanks!

src/Select.purs
JordanMartinez: Great! I was wanting to make those changes here, too.

> Also, I would prefer to name the input -> Input st something other than initialState and keep the Input st -> State st function named initialState. In this example I've renamed the new input -> Input st function simply mkInput, but you may have a better name for it.

Yeah, I can do that. For a better name, what about `incompleteState`? `mkInput` sounds good, but is `Input` really the best name to give for that now? It's no longer the input value to the component.

JordanMartinez:
> One last thing: do you mind removing the component template file for the time being and opening that as a separate PR? I'd like to consider it separate to this change. Thanks!

Sure. I'll have to get to that tomorrow though.

Also, should `Input st` be renamed to something else now that it's no longer the component's `input` type?

thomashoneyman:
> Sure. I'll have to get to that tomorrow though.

No rush!

> Also, should `Input st` be renamed to something else now that it's no longer the component's `input` type?

I've stuck with using the same type names as usual for the component -- `Message` even though you can replace the message wholesale with your own new type, `msg`, for example -- because as far as the Select component is concerned that still _is_ its message type (`H.raise` works on it), the `Input st` still is its input type (`initialState` works on it), and so on. It's just that the public interface can be changed to a different input or output (message).

So I'm leaning towards keeping things named as they are.

JordanMartinez:
> the `Input st` still is its input type (initialState works on it)

That's true for the `initialState` function, but is no longer true for the `receive` function (with the changes you requested above that I'm going to start working on)

**Edit:** Meaning, before we had this code:
```purescript
initialState :: Input st -> S.State st

receive :: Input st -> Maybe (S.Action action)
```
With our recent changes, now we have this code:
```purescript
mkInput :: input -> Input st

receive :: input -> Maybe action
```
From the end-users' perspective, `Input st` seems to be an odd name. I'd imagine users would expect something more like `IntermediateState st` since it tells them that this type isn't the full `State` type for a Select component.

<hr>

Regardless of whether we change `Input st` to a different name or not, I believe this PR is finished now.

JordanMartinez:
I've removed the component template file from this PR via a force push. The last two commits are new.

src/Select.purs
thomashoneyman: @davezuch do you have any opinions on the names here?

thomashoneyman: It's a bit confusing because we have various "inputs" going on:

1. The Select component requires a standard Halogen input type, currently called `Input st`
2. An extended Select component can expose a different input type to its users, currently represented by the type parameter `input`
3. The Select component produces its initial state from the `Input st` type as in usual Halogen, a function currently called `initialState`
4. An extended Select component requires a function to transform the `input` type into Select's `Input st` type, currently called `mkInput`.

So there's `Input` (for Select) and `input` (for extended Select); and there's `initialState :: Input -> State` (for Select) and `mkInput :: input -> Input st` (for extended Select).

Does this seem understandable, or does this bit of text bring to mind more apt names & types?

thomashoneyman:
I'm happy to merge this once we've settled the naming discussion. Thanks for spending your time on this! It's a nice change, not least because it allows you to hide the scary `Spec` type.

JordanMartinez:
I'd like to do this same change in your `formless` library. So, I assume the name discussion here will set a precedent for that library's PR.

src/Select.purs
JordanMartinez: Here's my proposal. These numbers don't correspond to Thomas' list above. They simply indicate how many things need to change:

1. Rename `Input st` to `IntermediateState st` as that's really what it is: an intermediate state value that is used to produce the component's actual initial state
2. Rename `mkInput` to `mkIntermediateState` to match the type change
3. Rename the Select component's `initialState` to `completeInitialState` to make the composition of these two functions sensible: `completeInitialState <<< mkIntermediateState`

As I was typing this comment, I realized it might make sense to rename `InputType` to `ComponentType` 

```purescript
{- old version:
data InputType = Text | Toggle                                       -}
data ComponentType = Text | Toggle

-- old version:
-- type Input st =
type IntermediateState st =
-- old version:
--{ inputType :: InputType
  { componentType :: ComponentType
  , search :: Maybe String
  , debounceTime :: Maybe Milliseconds
  , getItemCount :: {| st } -> Int
  | st
  }

-- proposed version only shown here
component :: MassivelyLongTypeSignatureIgnoredHere
component mkIntermediateState spec = H.mkComponent
  { initialState: completeInitialState <<< mkIntermediateState
  , -- render code 
  , -- eval code
  }
  where
  completeInitialState :: Input st -> State st
  completeInitialState = Builder.build pipeline
    -- rest of code
```

thomashoneyman:
Yes, the same would be true there. And great, I look forward to it!

src/Select.purs
thomashoneyman: This conversation is relevant over in Formless:
https://github.com/thomashoneyman/purescript-halogen-formless/pull/47#issuecomment-495710350

I'm hesitant to move to intermediateState because without extension the `Input` type as it stands really _is_ the component input, the same way the `Message` type really _is_ the message output without extension. There will definitely be cases in which you won't need to change the Select input type and could keep it as-is; perhaps less often than changing it, but I still don't think 'intermediate state' quite captures it. 

For example, if you didn't change the input type, then your component's input type would be `IntermediateState`.

JordanMartinez: I respond to both the "`input` is the component input when no extension occurs" and "`message` is the output when no extension occurs" concerns here:
- input concern: https://github.com/thomashoneyman/purescript-halogen-formless/pull/47#issuecomment-495744919
- message concern: https://github.com/thomashoneyman/purescript-halogen-formless/pull/47#issuecomment-495747988

JordanMartinez: ~Another idea that came to mind: we could remove the `Input st` type completely and change the type of `mkInput` to inline the definition of the `Input st` type.~
```purescript
-- current
mkInput :: input -> Input st

-- proposal
mkInput :: input -> 
  { inputType :: InputType
  , search :: Maybe String
  , debounceTime :: Maybe Milliseconds
  , getItemCount :: {| st } -> Int
  | st
  }
```
~However, I'd assume this would be bad for documentation purposes.~

**Edit:** Not just bad for documentation purposes. See the next comment for more issues.

JordanMartinez: ~And yet another idea (sorry for the multiple posts). @thomashoneyman Perhaps we can both get behind this one:~
```purescript
type MakeInput input st =
  input -> { inputType :: InputType
           , search :: Maybe String
           , debounceTime :: Maybe Milliseconds
           , getItemCount :: {| st } -> Int
           | st
           }

component :: MakeInput input st -> Spec -> Component -- rest of type signature
```

**Edit:** I tried this out in a separate branch. It doesn't work well in a few ways:
- Anytime one wants to specify what the unnamed `Input st` type is when they're defining the `mkInput` function for a Select/Formless component, they have to use the full type signature
- Formless does a few things with the resulting `Input` value before its `initialState` returns
- This idea isn't "future proof": if the definition changes, code still largely compiles since the type alias can be used rather than needing to update this everywhere in a code base.

thomashoneyman (approved):
I'm happy with these changes -- we can always consider different names later as there is not yet an official Halogen 5 release yet. I'll leave this PR active through the weekend to let the CitizenNet team get a look at it, and if they have no comments, I'll go ahead and merge.

Thanks!

JordanMartinez:
Sounds good!

At the very least, we could always not rename things and just explain this in the docs and/or tutorial for it.

davezuch (approved):
Sorry it took me so long to review this. I just got caught up on the conversation and commits. This all looks good to me! Thanks for making the PR. I'm happy with what's here.

Re: naming, `mkInput` feels more idiomatic and it's also clearer to me that it has to do with the component input. As @thomashoneyman pointed out, the other proposed names drop the "input" from the name which I think is important for communicating that function/type. So I'm in favor of keeping it as is.

I did have one question, what's the context for changing `Message` to `Event`?

thomashoneyman:
@JordanMartinez and I discussed this offline at the PureScript conference in LA the other day. Specifically, Jordan noticed that the `Message` type used in Select is no longer really accurate to what the type represents. While it's _possible_ to use the `Message` type as the component's output, it's not the default output and I'd expect most extended components to provide their own output type. It's only the output type if you opt-out of providing your own and use `H.raise` as your `handleMessage` function.

It's more accurate to think of it as an internal event you might want to hook in to. "A new search has been performed -- what do you want to do?" It's kind of like being subscribed to an event source and providing a handler more than being the component's output type. You might handle an event by raising it as a message, or you might insert some computation of your own, or whatever.

I liked the name change to `Event` because I think it opens the door for a little more creativity. You're no longer thinking of the type as "What messages do I want to pass to a parent component?" but rather "What events do I want users of this component to be able to hook in to?" There's much less reason to be conservative about the number of events because it's not going to spam the parent with unnecessary messages. Extended components can choose how to mediate the events.

davezuch:
Ah that makes sense. Works for me! I'll go ahead and merge this in.

JordanMartinez:
@thomashoneyman Thanks for summarizing that!

PR #49, opened 2019-11-28
Add helper types and template file

### LINKED ISSUE #46: Add convenience Halogen type aliases for hiding Select types
Basically, doing what `Formless` does for their [`Component` type](https://github.com/thomashoneyman/purescript-halogen-formless/blob/halogen-5/src/Formless/Types/Component.purs#L101).

For `Select`, this would be:
```purescript
type Component query slots input msg m =
  H.Component HH.HTML (Query query slots) input msg m

type ComponentHTML action slots m =
  H.ComponentHTML (Action action) slots m

type HalogenM st action slots msg m a =
  H.HalogenM (State st) (Action action) slots msg m Unit
```

I could also add type aliases that cover situations where one does not need to extend the component as Formless does (e.g. [`Component'`](https://github.com/thomashoneyman/purescript-halogen-formless/blob/halogen-5/src/Formless/Types/Component.purs#L105)), but I don't quite see the point of that if I add a component template file for Select

### LINKED ISSUE #47: Add template file for Select?
Similar to the ComponentTemplate.purs file I created in `learn-halogen` and what I added but later removed in #45 

I think this would be easier/better than #42 / #40

JordanMartinez:

Fixes #47 and fixes #46

## Other Notes:

While the examples would need to be updated somewhat, the template file itself already provides people with a compile-able example that they can adjust as needed.

I should probably specify a license for the template file so that others can knowingly modify it without concern of copyright issues.

### DIFF
--- a/src/Select.purs
+++ b/src/Select.purs
@@ -115,28 +115,37 @@ type Input st =
   | st
   }
 
+type Component query slots input msg m =
+  H.Component HH.HTML (Query query slots) input msg m
+
+type ComponentHTML action slots m =
+  H.ComponentHTML (Action action) slots m
+
+type HalogenM st action slots msg m a =
+  H.HalogenM (State st) (Action action) slots msg m a
+
 type Spec st query action slots input msg m =
   { -- usual Halogen component spec
     render
       :: State st
-      -> H.ComponentHTML (Action action) slots m
+      -> ComponentHTML action slots m
 
     -- handle additional actions provided to the component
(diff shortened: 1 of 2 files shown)


thomashoneyman (approved):

davezuch (approved):
Sorry for the late review, this looks good to me, thanks for submitting!

JordanMartinez:
Thanks!

PR #50, opened 2020-01-13
Switch to using KeyboardEvent.key instead of KeyboardEvent.code

skress:

Keyboard handling does not work in MS IE 11 and MS Edge as both browsers return "undefined" for `KeyboardEvent.code`. This PR changes the code to use `KeyboardEvent.key` instead.

In addition IE 11 returns "Up"/"Down"/"Esc" instead of "ArrowUp"/"ArrowDown"/"Escape".

To see it for yourself you can use https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code - there is a "Try it out" section where KeyboardEvent's key and code are displayed when pressing a key.

The PR only changes a couple of lines, so the review should be straightforward.

Just run the examples pre-PR and test keyboard handling in IE11/Edge. Then apply the PR and re-test.

### DIFF
--- a/src/Select.purs
+++ b/src/Select.purs
@@ -296,12 +296,12 @@ handleAction handleAction' handleEvent = case _ of
   Key ev -> do
     void $ H.fork $ handle $ SetVisibility On
     let preventIt = H.liftEffect $ preventDefault $ KE.toEvent ev
-    case KE.code ev of
-      "ArrowUp" ->
+    case KE.key ev of
+      x | x == "ArrowUp" || x == "Up" ->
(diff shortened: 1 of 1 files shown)


thomashoneyman (approved):
Thanks for finding and fixing this! The change works for me and I'm happy to merge. I'll leave this open for a day for the CitizenNet team to take a look at, and if there's no dissension then I'll merge it.

thomashoneyman:
Available at [release v5.0.0-rc.4](https://github.com/citizennet/purescript-halogen-select/releases/tag/v5.0.0-rc.4). Thanks!

PR #53, opened 2020-04-02
Migrate library to use Halogen Hooks

JordanMartinez:

Migrates Halogen Select to use the new Halogen Hooks library.

I'm not yet finished cleaning this PR up (renaming things and deleting the non-hook versions), but it gives you an idea on where I'm going.

## Miscellaneous Comments

A few comments/questions:
- Hooks don't have actions. Rather, their actions are `HookM slots output m Unit`. It's the only way to make hooks extensible without creating an `Action` type that wraps another `Action` type.
- I'm not sure whether the state token should be exposed or not. My concern is inconsistent state in the underlying hook, but that concern might not be valid.
- While it's easier to write the render and action portions of a component via `let` because everything is in scope, it's easier to read the resulting code via `where`. This code should likely be refactored to use `where`.

## General Hook Pattern

After implementing this, I think this is the pattern to follow:
```purescript
type RenderlessComponentReturn =
 { state :: RenderlessComponentState
 , props :: RenderlessComponentProps
 , queries :: RenderlessComponentQueries
 }

selectComponent = Hooks.component \input -> Hooks.do
 -- renderlessComponentName <- useRenderlessComponentName
 select <- useSelect -- args...
  when (select.state.highlightedIndex == ..) do
  result <- select.queries.getSomeValue arg1
  Hooks.pure $ HH.div [ select.props.container <> [ class_ "my-class-name" ]
```


thomashoneyman:
Thanks for getting this started, @JordanMartinez!

> Hooks don’t have actions

Right, with Hooks you won’t write an action type. Instead you’ll allow anything that runs in `HookM` so long as the child slots, output messages, and monad match. So, for example, you can just write regular `ComponentHTML`  and call the props functions on that without having to worry about which context you’re in (is this my action, or Select’s?).

> I’m not sure whether the state token should be exposed

I don’t think it’s a good idea to expose the state tokens. I don’t think users should have unrestricted access to read and modify Select’s internal state. This wasn’t possible to do with Renderless.

The type you provided for what Select returns looks right. If I’m on the same page as you, then we agree that Select should return to you:

1. The values in state that matter (highlighted index, visibility, etc.), while hiding any values meant for internal use
2. Functions for updating the values in state that ought to be modifiable (we may want read-only values in state). These functions should not expose the state token.
3. Functions that should be used to set props on particular parts of the UI, like the input field or the container.

#### 1. State
It looks like all the values in Select’s state currently matter, so we can return all of them to the user. Still, I’d like to make sure future changes to the library keep in mind that Select should be _intentional_ about what information it returns to the user — not just the whole state by default.

#### 2. Functions on State
We shouldn’t allow unfettered access to act on Select’s state. Instead, we should return a set of functions the user can use.

This is conceptually similar to returning actions the user can embed in their HTML or extend with their own action type, but it’s more convenient because these functions and the user’s actions would be the same type.

It’s actually more similar to writing a query interface because these functions aren’t just actions — they can also return values. So you could use them in any `HookM` block to return values from state. But they’re WAY more convenient than queries.

This is nice because we get the power of queries without the overhead of child components or having to make things extensible (with the corresponding complexity and tedium). We also get to control what values in Select should be read-only — you can only act on Select’s state using functions we provide.

#### 3. Props Setters
Props setters are pretty much unchanged, but they can be returned by the Hook directly instead of imported. That way we can just bake the state directly in to these setters; the user doesn’t have to ever provide it as an argument (nor do they actually ever see the internal state directly).

Right now we don't even seem to pass the state in to these, so maybe they don't have to returned by the Hook and can just be functions in module scope.

#### An aside: `handleEvent`
The weird case is `handleEvent`. This function needs access to the actions described by point 2 above, but when the user implements this function they won’t yet have access to those actions.

So instead of having this be:

```purs
handleEvent :: StateToken SelectState -> Event -> HookM
```

We could have it be

```purs
handleEvent :: { focus :: HookM, setVisibility :: HookM, ... }
handleEvent :: Actions -> Event -> HookM
```

and Select partially-applies this handler with the actions it also returns to the user in the Hook.

##  Handling future improvements
Hooks make it convenient to add even more to Select without increasing cognitive overhead as an implementor or user. If we return state, functions on state (we could still refer to them as ‘actions’), and props setters, then it’s easy to extend from there.

For example, let’s say we wanted to introduce new functionality to Select: the ability to reset the component state. We would:

1. Add a new field to state to store the initial component state — but not return this in the Hook, keeping this information internal
2. Add a new ‘action’ called `reset` that, internally, reads the initial state and puts it as the new state

The end user gets a new field in the hook’s response, `reset`, which they can freely ignore via destructuring.

Any further improvements to Select would follow this workflow: update state, update actions if necessary, add to props setters internals’ if necessary.

JordanMartinez:
So, I've run into a problem. When passing the `inputRec` into the `useSelect` hook, I have to specify what `handleEvent` is. The Dropdown example uses the `Selected` event to set the visibility off. The problem is cyclical dependencies: I can't call `select.setSelected Off` because `useSelect` hasn't returned that action yet.

I think the solution is to do FRP via `purescript-event` or something. If I can "push" the change notification into some event stream, the end-user can handle such changes when they come through the pipe. In such a situation, the end-user will have `select.setSelected` accessible.

JordanMartinez:
Also, my goal was first getting everything to compile (which it does now). I'm now trying to clean things up and refactor this code to account for the things you mentioned above. I was hoping to do this piecemeal, so one could see the logic behind what I'm doing. However issues like what I describe above make that problematic.

thomashoneyman:
Does this not work? (The section titled "An aside: `handleEvent`" above)

> The weird case is handleEvent. This function needs access to the actions described by point 2 above, but when the user implements this function they won’t yet have access to those actions.

In that I thought `handleEvent` could take the actions it needs access to as its first argument. The user's implementation of the function should be written using those actions. Then, Select will provide those actions internally (the same actions it also returns in the hook).

So the type of `handleEvent` would be:

```purs
handleEvent :: { focus :: HookM, setVisibility :: HookM, ... }
handleEvent :: Actions -> Event -> HookM
```

...and internally, Select would call something like:

```
Hooks.do
  ...
  let
    handleEvent' :: Event -> HookM _ _ _ Unit
    handleEvent' = input.handleEvent { focus, setVisibility, ... }

  Hooks.pure
    { ...state
    , focus
    , setVisibility
    }
```

where the same actions applied to the user's `handleEvent` are the ones returned by the Select hook. It's possible there's a nicer way to do this, but this seems like it would work; I haven't tested it out, though.

JordanMartinez:
Ah... Yeah, that could work. Maybe not the nicest, but it'll do.

JordanMartinez:
Doh! 
```purescript
    let
      setVisibility :: Visibility -> HookM slots output m Unit
      setVisibility v = do
        st <- Hooks.get stateToken
        when (st.visibility /= v) do
          Hooks.modify_ stateToken (_ { visibility = v, highlightedIndex = Just 0 })

          -- eventAPI is needed in setVisibility's definition
          inputRec.handleEvent eventAPI $ VisibilityChanged v

      eventAPI :: EventAPI slots output m
      eventAPI = { setFocus, setVisibility }
```
I could refactor this out so that only `setFocus` is passed into here. However, that won't scale.
Still, even if I went with the FRP approach, one could create an infinite loop by setting the visibility to the same value when notified that the visibility has changed.

JordanMartinez:
The above idea I had about only passing `setFocus` only works because there is not a corresponding `FocusChanged` event that gets handled. If there was, then we'd have a cycle in our dependencies. `setFocus` would not be able to run unless it could use `setVisibility`. Likewise, `setVisibility` wouldn't be able to run without being able to use `setFocus`.

Speaking to Thomas on Slack, he asked me to try the FRP approach in a separate branch.

JordanMartinez:
So, the issue I run into is that `purescript-event`'s `subscribe` function requires the callback to be an `Effect`. In other words, I can't use `HookM` there to run actions or set state.

I'm wondering whether this is a situation where `purescript-coroutines` might work...? Idk. I'm probably too tired to think straight. :sleeping_bed:

JordanMartinez:
## Summary of the Actions Problem so far: Cyclical Dependencies

1. The problem we're trying to solve (more or less) is that when an event occurs inside the hook (e.g. visibility changed), the end-user can handle such events (e.g. no longer display X). By migrating this library to the Hooks approach without changing anything else, the same functionality can be provided if we expose the state tokens for the Select hooks' internal state. The end-user can adjust the internal state as desired.
2. The problem with exposing the hook's state tokens is that the end-user can modify the internal state in such a way that the state becomes inconsistent. For example, one cannot just set the state value for visibility to `true`. One also needs to focus the corresponding HTML element. The solution to this is to return via `Hooks.pure` "actions" that can modify the hook's internal state correctly (e.g. `setFocus` and `setVisibility`). 
3. The next problem is that the "actions" themselves emit events that the end-user should be able to handle. For example, calling `setFocus` will update the visibility of the element, but it will also create a `VisibilityChanged` event in the hook. It's this last problem that creates a cyclical dependency. To define `setFocus`, the hook needs to know how to handle the `VisibilityChanged` event. Right now, this is passed in as an argument to the hook via `inputRec`. To handle the event, the end-user needs access to the "actions" returned by the hook.
4. In Jordan's attempt to use FRP as a way to solve this, `purescript-event`'s `subscribe` function requires the event handler to run in the `Effect` monad, not the `HookM` monad.

## Principles for Solutions

Here's some principles to help guide our thoughts on possible solutions
1.The moment the event occurs is the moment it is handled. In other words, we shouldn't say "Event X occurred," wait 4 seconds, and then handle event X. Otherwise, we might persist state that is stale.
2. The end-user needs full-access to the hook's actions when handling the event. This implies that the hook needs to also return some "event listener"-like thing.
3. The end-user should not "pay" for any overhead if they choose not to listen to such events. Just because the functionality exists doesn't mean we need to use it in every case.
4. The end-user should have control as to _when_ to subscribe/unsubscribe to such events. For example, subscribe on initialization and unsubscribe on finalization. Or subscribe when X becomes true and unsubscribe when X becomes false. Or any combination of these two.
5. The event handler needs to produce a `Hook` that takes a `a -> HookM` callback. (e.g. `select.onNewSearch \ix -> do ...`
6. The listener/producer does not require an initial/default value.

I1. (Ideal, but not pressing for right now) the end-user can subscribe to the same event multiple times and set up composable pipelines that handle events in a custom way (e.g. when event X occurs and event Y happened 3 seconds go, then do Z).

## Candidates

- `purescript-event`: does not satisfy 5 because the callback returns `Effect`, not `HookM`.
- `purescript-coroutine`: does not satisfy 6.
- `purescript-avar`: haven't tested this yet

JordanMartinez:
I figured out a solution. I'll work from the bottom up.

## Introducing the main idea

When we use `Hooks.captures {} Hooks.useTickEffect someHookMCode`, we can run `someHookMCode` once per render. When we use it with some memos, it only runs when those values change (e.g. `Hooks.captures { state } Hooks.useTickEffect someHookMCode`. Thus, I came up with this general idea:
```purescript
useEvent
  :: forall slots output m a 
   . Eq a
  => Hook slots output m (UseEvent a) _
useEvent = Hooks.wrap Hooks.do
  state /\ tState <- useState Nothing

  Hooks.pure 
             -- whenever we push a new value, we update the state
             { push: \value -> Hooks.put tState (Just value)

             -- which will trigger `useTickEffect` because the memo has changed
             , subscribe: \callback -> Hooks.captures { state } Hooks.useTickEffect do
                 state' <- Hooks.get tState
                 for_ state' cb
                 pure Nothing -- no need to unsubscribe later
             }
```
The problem with this implementation is that you get a compiler error: "the type variable `x`, bound at `<some location>` has escaped its scope, appearing in the type, `y`"

The issue seems to be caused by `useTickEffect`. Since we're using Indexed Monads, we're calling `useTickEffect` in a location other than its actual usage (e.g. in `DropdownHook`). So, the workaround is to redefine the function, so that calling `useTickEffect` is done by the user of it. That changes the function's implementation to this:
```purescript
useEvent
  :: forall slots output m a 
   . Eq a
  => Hook slots output m (UseEvent a) _
useEvent = Hooks.wrap Hooks.do
  state /\ tState <- useState Nothing

  Hooks.pure 
             { push: \value -> Hooks.put tState (Just value)
             , captures: Hooks.captures { state }
             , subscribe: \cb -> do
                 state' <- Hooks.get tState
                 for_ state' cb
                 pure Nothing
             }
```
In `SelectHook`, we can create and use it like this:
```purescript
useSelect :: _
useSelect :: Hooks.wrap Hooks.do
  onSelectedIdxChanged <- useEvent

  -- somewhere else
    onSelectedIdxChanged.push 4

  Hooks.pure 
    { state
    , onSelectedIdxChanged: 
        { captures: onSelectedIdxChanged.captures
        , subscribe: onSelectedIdxChanged.subscribe
        }
    }
```
Now, we can use it like this:
```purescript
dropdownExample :: _
dropdownExample = Hooks.component \_ -> Hooks.do
  select <- useSelect inputRec
  select.onSelectedIdxChanged.captures Hooks.useTickEffect do
    select.onSelectedIdxChanged.subscribe \ix -> do
      -- rest of the function here...
```

That's the basic idea. 

## Cleaning up the API

To make it easier to return the `captures` and `subscribe` functions, we wrap them in a `props` field instead:
```purescript
useEvent
  :: forall slots output m a 
   . Eq a
  => Hook slots output m (UseEvent a) _
useEvent = Hooks.wrap Hooks.do
  state /\ tState <- useState Nothing

  Hooks.pure 
             { push: \value -> Hooks.put tState (Just value)
             , props: { captures: Hooks.captures { state }
                      , subscribe: \cb -> do
                          state' <- Hooks.get tState
                          for_ state' cb
                          pure Nothing
                      }
             }

-- in useSelect
  Hooks.pure 
    { state
    , onSelectedIdxChanged: onSelectedIdxChanged.props
    }
```

However, notice how we're using the `Eq a` constraint in the `useEvent` type signature. That's because `Hooks.captures` uses that. What if our type didn't have an `Eq` instance or we wanted to use a different way to determine equality? In that situation, we need to use `Hooks.capturesWith` instead. That would change our code to the following:
```purescript
useEvent
  :: forall slots output m a
   . Hook slots output m (UseEvent a) _
useEvent = Hooks.wrap Hooks.do
  state /\ tState <- useState Nothing

  Hooks.pure { push: \value -> Hooks.put tState (Just value)
             , props: { capturesWith: \eqFn -> Hooks.capturesWith eqFn { state }
                      , subscribe: \cb -> do
                          state' <- Hooks.get tState
                          for_ state' cb
                          pure Nothing
                      }
             }
```
... and in the `DropdownHook` we would now add the `(==)` after the `capturesWith` function:
```purescript
select <- useSelect inputRec
select.onSelectedIdxChanged.capturesWith (==) Hooks.useTickEffect do
  select.onSelectedIdxChanged.subscribe \ix -> do
    -- handler
```
I'm currently trying to figure out how to clean up the API further, so that users of the library only have to write:
```purescript
  select.onSelectedIdxChanged (==) Hooks.useTickEffect \ix do
   -- handler
```

JordanMartinez:
I just experienced type signature hell. It's time to go to bed.

In short, I was trying to create a function like this:
```purescript
subscribeTo :: _
subscribeTo props eqFn cb =
 props.captureWith eqFn Hooks.useTickEffect do
  props.subscribe cb
  pure Nothing

-- example usage
subscribeTo select.onSelectedIdxChanged (==) \ix -> do
  -- handler
```
I used the type signatures specified by the compiler via typed holes, tried to clean them up with readable names, and it would always fail somewhere along the line. So... I just copied and pasted what the compiler got. Sigh...

See [this branch](https://github.com/JordanMartinez/purescript-halogen-select/tree/useStateAndTick)

JordanMartinez:
Ok. I cleaned up the type signatures after turning on spago's `watch` feature to get a faster feedback loop (IDE was untrustworthy). This is what we write now:
```purescript
-- SelectHook
useSelect = Hooks.wrap Hooks.do
  onSelectedIdxChanged <- useEvent

  -- somewhere in code
  onSelectedIdxChanged.push 4

  Hooks.pure
    { state
    , setFocus
    , setVisibility
    , onSelectedIdxChanged: onSelectedIdxChanged.props
    , -- ...
    }

-- Dropdown Hook
Hooks.component \_ -> Hooks.do
  select <- useSelect inputRec
  subscribeTo select.onSelectdIdxChanged \ix -> do
    -- handler
```
and similarly for the Typeahead example

JordanMartinez:
Latest issue I've run into. I've implemented `useEvent` in the following manner:
```purescript
useEvent
  :: forall output m slots a
   . Hook slots output m (UseEvent a) (EventApi slots output m a)
useEvent = Hooks.wrap Hooks.do
  state /\ tState <- useState Nothing

  Hooks.pure { push: \value -> Hooks.put tState (Just value)
             , props: { capturesWith: \eqFn -> Hooks.capturesWith eqFn { state }
                      , subscribe: \cb -> do
                          state' <- Hooks.get tState
                          for_ state' cb
                          Hooks.put tState Nothing
                      }
             }
```
There are two issues with this:
1. Every time a value is pushed, the hook will rerender, even if no one subscribes to the event. This could cause an unneeded performance issue if one uses a lot of events
2. To make it so that the next pushed value always causes the callback to be run, the state is modified a second time to `Nothing`. Since all pushed values will be `Just value`, `Nothing` will never equal `Just _` and cause the callback to be run.

The first could potentially be fixed by storing the callback in a `Ref` and only updating the state if the `Ref` stores something. When one subscribes to events, 
I'm not sure about the second yet.

JordanMartinez:
@thomashoneyman Besides updating documentation, I think this is ready to go. The code is much cleaner now.

JordanMartinez:
AFAICT, this PR is mostly good. However, see jordanmartinez/purescript-halogen-hooks-extra#1 and jordanmartinez/purescript-halogen-hooks-extra#2

thomashoneyman:
Superseded by #54

PR #54, opened 2020-05-04
[Take 2] Migrate library to use Halogen Hooks

JordanMartinez:
Uses `halogen-hooks` `v0.4.0` and `halogen-hooks-extra` `v0.6.0`.

See examples files for what the end interface looks like.


JordanMartinez:
Once a new package set with `v0.5.0` of halogen-hooks-extra comes out, I think this PR can be merged. Right now, I'm overriding that package in the set.

src/Select.purs
```diff
@@ -90,266 +78,317 @@ derive instance ordVisibility :: Ord Visibility
...
-  , getItemCount :: {| st } -> Int
-  | st
   }
 
-type Debouncer =
-  { var :: AVar Unit
-  , fiber :: Fiber Unit
-  }
+type SelectReturn
+      m toggleProps itemProps containerProps inputProps =
```
thomashoneyman: These don't feel like they'd be necessary in the return type. Can these type variables be pushed down into the props functions themselves?

```purs
toggleProps :: forall props. Array (HP.IProp (ToggleProps props) (HookM m Unit))
```

The monad type `m` would need to remain in the return type, though, to ensure all the `HookM m ...` unify.

I'm assuming you didn't want to put these in the type in the first place and ran into some type errors. If so, what happens if you `newtype` the Select return type? I just have the sense that this shouldn't be necessary.

src/Select.purs
```diff
@@ -90,266 +78,317 @@ derive instance ordVisibility :: Ord Visibility
...
+-- | select <- useSelect $ selectInput
+-- |   { getItemCount = pure (length items)
+-- |   , pushNewSearch = events.push
+-- |   }
+-- | ```
+selectInput :: forall m. SelectInput m
+selectInput = do
+  let
+    ignoreEvent :: forall a. a -> HookM m Unit
+    ignoreEvent = \_ -> pure unit
```
thomashoneyman: This can also be defined as `mempty`, which simplifies your usage below:

```purs
, pushNewSearch: mempty
, pushVisibilityChanged: mempty
, ...
```

src/Select.purs
```diff
@@ -90,266 +78,317 @@ derive instance ordVisibility :: Ord Visibility
...
+  , getItemCount: pure 0
+  , pushNewSearch: ignoreEvent
+  , pushVisibilityChanged: ignoreEvent
+  , pushSelectedIdxChanged: ignoreEvent
   }
 
-component
-  :: forall st query action slots input msg m
+useSelect
+  :: forall m toggleProps itemProps containerProps inputProps
```
thomashoneyman: If you're able to remove these from the return type then they can be removed here, too.

src/Select.purs
```diff
@@ -90,266 +78,317 @@ derive instance ordVisibility :: Ord Visibility
...
+        -- key events and expire their search after a set number of milliseconds.
+        _ -> pure unit
+
+    Hooks.pure
+      -- state
+      { search: state.search
+      , visibility: state.visibility
+      , highlightedIndex: state.highlightedIndex
+
+      -- actions
```
thomashoneyman, line 237: Since there are only three actions being returned to the user, and the others are being used to set props, it seems like some refactoring could be done to reduce all the argument passing happening here.

Are you open to moving functions that depend on `state` up into let bindings here, to skip the argument passing, while leaving others which don't depend on it directly (like `setFocus`) in the where clause?

As a rule of thumb, anything that would take `state` or `modifyState` as an argument can be moved up into a let binding, and everything else can remain in the  `where` clause.

That way, once you've defined the actions that depend on state in `let`s you can just pass the resulting `HookM m Unit` functions directly to things like the `toggleProps` function, which reduces all the coordination being done among these various functions.

Still others can just be defined inline -- for example, `clearSearch` might as well just be defined inline as `clearSearch: modifyState _ { search = "" }` 

So, for example:

```purs
let
  setVisibility :: Visibility -> HookM m Unit
  setVisibility v = do
    when (state.visibility /= v) do
      modifyState (_ { visibility = v, highlightedIndex = Just 0 })
      inputRec.pushVisibilityChanged v

Hooks.pure
  { setVisibility
  , clearSearch: modifyState _ { search = "" }
  , toggleProps: toggleProps setVisibility toggleClick key
  }
where
toggleProps :: (Visibility -> HookM m Unit) -> ...
```

src/Select.purs
```diff
@@ -90,266 +78,317 @@ derive instance ordVisibility :: Ord Visibility
...
+          true -> HTMLElement.focus el
+          _ -> HTMLElement.blur el
+
+      setVisibility
+        :: HookM m SelectState
+        -> ((SelectState -> SelectState) -> HookM m Unit)
+        -> (Visibility -> HookM m Unit)
+        -> Visibility
+        -> HookM m Unit
+      setVisibility getState modifyState pushVisibilityChanged v = do
```
thomashoneyman: By moving these implementations into `let` you can skip most of these function arguments and avoid using `getState` altogether.

src/Select.purs
```diff
@@ -90,266 +78,317 @@ derive instance ordVisibility :: Ord Visibility
...
+            setVisibility getState modifyState pushVisibilityChanged On
+
+      key
+        :: HookM m SelectState
+        -> ((SelectState -> SelectState) -> HookM m Unit)
+        -> (Visibility -> HookM m Unit)
+        -> (Int -> HookM m Unit)
+        -> KeyboardEvent
+        -> HookM m Unit
+      key getState modifyState pushVisibilityChanged pushSelectedIdxChanged ev = do
```
thomashoneyman: Even if you don't decide to move any of these into `let` bindings you can still refer to the input directly, without passing it as an argument. For example, both of these `push*` functions are already available via `inputRec` which is in scope.

src/Select.purs
JordanMartinez: Yeah... I ran into "`slot` type escaped its scope" errors. Now that the `slot` type is no longer a part of the `HookM` definition, I might be able to make this change.

src/Select.purs
JordanMartinez: Good point. I'll update that.

src/Select.purs
JordanMartinez: Nope. It still happens. 
```
[1/1 EscapedSkolem] src/Select.purs:173:6

            v
  173    in Hooks.wrap Hooks.do
  174      state /\ modifyState <- useState
  175        { search: initialSearchValue
       ...
  205        , containerProps
  206        , inputProps: inputProps getState modifyState searchDebouncer inputRec.pushVisibilityChanged inputRec.pushSelectedIdxChanged
  207        }
             ^
  
  The type variable toggleProps, bound at
  
    src/Select.purs:106:20 - 106:97 (line 106, column 20 - line 106, column 97)
  
  has escaped its scope, appearing in the type
  
    Hooked m0 hooks1
      (UseDebouncer String
         (UseGet
            { highlightedIndex :: ...
            , search :: String
            , visibility :: Visibility
            }
            (UseState ... hooks1)
         )
      )
      { clearSearch :: HookM m0 Unit
      , containerProps :: Array (IProp ... ...)
      , highlightedIndex :: Maybe Int
      , inputProps :: Array (IProp ... ...)
      , itemProps :: Int -> Array ...
      , search :: String
      , setFocus :: Boolean -> HookM m0 Unit
      , setVisibility :: Visibility -> HookM m0 Unit
      , toggleProps :: Array (IProp ... ...)
      , visibility :: Visibility
      }
  
  in the expression ((bind ixBindIndexed) (useState { search: ...
                                                    , visibility: ...
                                                    , highlightedIndex: ...
                                                    }
                                          )
                    )
                    (\$5 ->
                       case $5 of
                         (Tuple state modifyState) -> ...
                    )
  in value declaration useSelect
```

src/Select.purs
JordanMartinez: Oh, good point! I forgot about that.

src/Select.purs
JordanMartinez: I've inlined `clearSearch` in latest commits.

src/Select.purs
JordanMartinez: Done in latest commits.

src/Select.purs
JordanMartinez: Done in latest commits

src/Select.purs
JordanMartinez: I haven't done the other work yet because when I was thinking about this one, I [wondered whether `useGet` is necessary in case one wants to use a returned action in an Effect cleanup / forked computation](https://github.com/thomashoneyman/purescript-halogen-hooks/issues/30#issuecomment-627712157)

thomashoneyman: It's not necessary if you're modifying state (`modifyState` always uses fresh state), but it would be if you are relying on the `state` variable to say perform an effect in one of these actions. But then you have to basically guess whether it would ever make sense to do a particular action in an effect cleanup / forked computation. If the answer is yes, and you do rely on the `state` variable to perform an effect, then you should use `useGet`.

JordanMartinez: > It's not necessary if you're modifying state (`modifyState` always uses fresh state), but it would be if you are relying on the state variable to say perform an effect in one of these actions.

Another situation that can arise is if you need to refer to the state that a library returns. For example, if I had a hook like this:
```purescript
state' <- do
  state /\ stateId <- useState 0
  useTickEffect { } $ Hooks.modify_ stateId (_ + 1)
  pure state
useTickEffect { } do
  pure $ Just do
    liftEffect $ log $ "This value is stale: " <> show state
```
... the state value is stale. There are two ways this could be fixed
1. the library **user** uses `getState <- useGet state`

```purescript
state <- do
  state /\ stateId <- useState 0
  useTickEffect { } $ Hooks.modify_ stateId (_ + 1)
  pure state
getState <- useGet state
useTickEffect { } do
  pure $ Just do
    state' <- getState
    liftEffect $ log $ "This value is stale: " <> show state
    liftEffect $ log $ "This value is not stale: <> show state'
```

2. the library **author** returns `{ state, getState: Hooks.get stateId }`
```purescript
{ state, getState } <- do
  state /\ stateId <- useState 0
  useTickEffect { } $ Hooks.modify_ stateId (_ + 1)
  pure { state, getState: Hooks.get stateId }
useTickEffect { } do
  pure $ Just do
    state' <- getState
    liftEffect $ log $ "This value is stale: " <> show state
    liftEffect $ log $ "This value is not stale: <> show state'
```

src/Select.purs
thomashoneyman: Hmm. This still strikes me as odd. This has the record printed out -- did you try adding a `newtype` here? I really don't want to force folks who want to annotate the return type to have to specify what props these functions would receive as input.

src/Select.purs
thomashoneyman: With Hooks 0.4.0 this is back to the usual.

spago.dhall
```diff
@@ -0,0 +1,19 @@
...
+  , "effect"
+  , "halogen"
+  , "halogen-hooks"
+  , "halogen-hooks-extra"
+  , "psci-support"
+  , "record"
+  ]
+, packages = ./packages.dhall
+, sources = [ "src/**/*.purs", "test/**/*.purs" ]
+}
```
thomashoneyman, line 15: If you'd like to switch the project to Spago, I'd like to ensure that generating a new `bower.json` file from the Spago dependencies produces a Bower file that matches the dependencies and versions currently there. That probably means that you move the test sources to a separate `test.dhall` file, and don't include dependencies like `debug`, `console`, `argonaut`, and `psci-support` (perhaps those should go in an `examples.dhall` file?) The `halogen-hooks` and `halogen-hooks-extra` are OK.

You might want to look at the migration to Spago for Halogen as a comparison:
https://github.com/purescript-halogen/purescript-halogen/pull/653

src/Select.purs
JordanMartinez: I don't get what you mean by "adding a newtype here." So, I haven't tried doing that yet.

src/Select.purs
JordanMartinez: I'll update this PR to use that version later on today.

src/Select.purs
thomashoneyman: Sorry, I mean `newtype SelectReturn m = SelectReturn { ... }`

src/Select.purs
JordanMartinez: Yup, the newtype fixed it. See latest commits.

src/Select.purs
JordanMartinez: Done!

src/Select.purs
JordanMartinez: I decided not to do that. The number of args passed to the function are smaller now because of the `inputRec` being in scope. I did try putting these back into the `let` statement, but it still seemed verbose.

spago.dhall
JordanMartinez: How do I generate a bower.json file using Spago? It looks like `spago publish` is no longer a command in Spago `v0.15.2`.

Also, there is no `test` folder.

I've split up the dependencies between `spago.dhall` and `examples.dhall`

thomashoneyman: That’s odd — I used it in the Halogen migration and I can’t find anything about it in the release notes. Without that command I don’t know how to generate one.

JordanMartinez: D'oh! Looks like its `spago bump-version` that generates the `bower.json` file.

JordanMartinez: Looks like it only worked for you because `halogen` was already on the Bower registry before they stopped all future uploads of new packages. Since our packages weren't already on there, this fails:
```
$ spago bump-version major
[info] Generating a new Bower config using the package set versions..
[error] Failed to run: `bower info --json purescript-halogen-hooks-extra#v0.6.0`
[error] [{
  "code": "ENOTFOUND",
  "data": {
    "endpoint": {
      "name": "",
      "source": "purescript-halogen-hooks-extra",
      "target": "v0.6.0"
    }
  },
  "id": "ENOTFOUND",
  "level": "error",
  "message": "Package purescript-halogen-hooks-extra not found",
  "stacktrace": "Error: Package purescript-halogen-hooks-extra not found\n    at createError (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/util/createError.js:4:15)\n    at /home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/core/resolverFactory.js:243:23\n    at _fulfilled (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:854:54)\n    at /home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:883:30\n    at Promise.promise.promiseDispatch (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:816:13)\n    at /home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:624:44\n    at runSingle (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:137:13)\n    at flush (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:125:13)\n    at processTicksAndRejections (internal/process/task_queues.js:79:11)"
}]

[error] Failed to run: `bower info --json purescript-halogen-hooks#v0.4.0`
[error] [{
  "code": "ENOTFOUND",
  "data": {
    "endpoint": {
      "name": "",
      "source": "purescript-halogen-hooks",
      "target": "v0.4.0"
    }
  },
  "id": "ENOTFOUND",
  "level": "error",
  "message": "Package purescript-halogen-hooks not found",
  "stacktrace": "Error: Package purescript-halogen-hooks not found\n    at createError (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/util/createError.js:4:15)\n    at /home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/core/resolverFactory.js:243:23\n    at _fulfilled (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:854:54)\n    at /home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:883:30\n    at Promise.promise.promiseDispatch (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:816:13)\n    at /home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:624:44\n    at runSingle (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:137:13)\n    at flush (/home/jordan/.nvm/versions/node/v13.12.0/lib/node_modules/bower/lib/node_modules/q/q.js:125:13)\n    at processTicksAndRejections (internal/process/task_queues.js:79:11)"
}]
```

thomashoneyman: Oh, shoot. Hmm. What if we add an explicit GitHub URL for these instead of pulling from the package set? We can't publish the new library to Pursuit without a valid Bower file. In the worst case we can just ensure that the Spago.dhall dependencies match the Bower dependencies, at the same major verisons, before release.

src/Select.purs
```diff
@@ -90,266 +77,300 @@ derive instance ordVisibility :: Ord Visibility
...
-
-    -- optionally perform some action on initialization. disabled by default.
-  , finalize
-      :: Maybe action
-  }
+-- | When pushing all Select events into the same handler, this data type
+-- | distinguishes one event type from another.
+data SelectEvent
+  = NewSearch String
+  | VisibilityChangedTo Boolean
```
thomashoneyman: Can we keep this as `VisibilityChangedTo Visibility`?

src/Select.purs
```diff
@@ -90,266 +77,300 @@ derive instance ordVisibility :: Ord Visibility
...
+  , itemProps
+      :: forall itemProps
+       . Int
+      -> Array (HP.IProp (ItemProps itemProps) (HookM m Unit))
+  , containerProps
+      :: forall containerProps
+       . Array (HP.IProp (onMouseDown :: ME.MouseEvent | containerProps) (HookM m Unit))
+  , inputProps
+      :: forall inputProps
+       . Array (HP.IProp (InputProps inputProps) (HookM m Unit))
```
thomashoneyman, line 154: These were originally functions `Array _ -> Array _` so that `append` could be pre-applied. That was done because we want to make sure these props get overridden by the user's props, if they supply a duplicate. It's also a little more convenient at the call site -- `props <> [ ]` or `append props []` vs. `setProps []`.

What do you think of returning to that approach? Did you have a reason to move to this one instead?

src/Select.purs
```diff
@@ -90,266 +77,300 @@ derive instance ordVisibility :: Ord Visibility
...
+    state /\ stateId <- useState
+      { search: initialSearchValue
+      , visibility: Off
+      , highlightedIndex: Nothing
+      }
+
+    searchDebouncer <- useDebouncer debounceTime \lastSearchState -> Hooks.do
+      case inputRec.inputType of
+        Text -> do
+          -- modifyState (_ { highlightedIndex = (Just 0) })
```
thomashoneyman: Should this be uncommented?

thomashoneyman:
This is really close. I'm happy to merge it in to `master`, but probably not make a release until the README can be updated to be more useful. Users can try out the new Hooks version by updating their `spago.dhall` to have version = "master"` for the library.

I think we should let the library sit a little longer while updating others like Formless before formally releasing, so that we can continue making tweaks as needed.

spago.dhall
JordanMartinez: I mean... there's always [Starsuit](https://spacchetti.github.io/starsuit/)

>  In the worst case we can just ensure that the Spago.dhall dependencies match the Bower dependencies, at the same major verisons, before release.

Meaning the `bower.json` file matches the depdencies' versions specified in the package set?

src/Select.purs
JordanMartinez: Yup!

src/Select.purs
JordanMartinez: Personally, I thought `props` is better than `setProps []`. If a user wants to override them, wouldn't `props <> [ override1Prop ]` work, too?

src/Select.purs
JordanMartinez: Nope! I swtiched `modifyState` to `H.modify_` and it took a while to realize I needed `Hooks.modify_` not the other one. I likely commented this out, thinking there was another issue elsewhere and then forgot to uncomment it when I found out what the real problem was.

JordanMartinez:
> I think we should let the library sit a little longer while updating others like Formless before formally releasing, so that we can continue making tweaks as needed.

Agreed!

src/Select.purs
JordanMartinez: Fixed in latest commits.

src/Select.purs
JordanMartinez: Fixed in latest commits.

src/Select.purs
thomashoneyman: It does work, it’s just that it’s really easy to write it the other way around and not know why your override isn’t working

JordanMartinez: Ah.... that makes sense. And if you, for whatever reason, decided to combine multiple libraries' props together, you'd get
```purescript
-- no overrides
setSelectProps (setFormlesProps (setCustomHookProps []))

-- with overrides
setSelectProps (setFormlesProps (setCustomHookProps 
  [ overrideSelectProp
  , overrideFormlessProp
  , overrideCustomHookProp
  ]))
```
and even the order in which you 'set' the props matters because Formless could override Select's corresponding prop.

I'll make the change back to `setProps`. If we want to approach this in some other manner, we can in a future PR.

JordanMartinez: Done in latest commit.

examples/Components/Dropdown.purs
```diff
@@ -2,24 +2,22 @@ module Components.Dropdown where
...
-
-type Slot =
-  H.Slot S.Query' Message
 
-type State =
-  ( items :: Array String
-  , selection :: Maybe String
-  , buttonLabel :: String
-  )
+type Slot query = H.Slot query Message
```
thomashoneyman: This is `Const Void` below for the query type

```suggestion
type Slot = H.Slot (Const Void) Message
```

examples/Components/Typeahead.purs
```diff
@@ -44,97 +38,90 @@ data Message
...
-  -- input instead of the usual select one.
-  input :: S.Input State
-  input =
-    { inputType: S.Text
-    , debounceTime: Just (Milliseconds 300.0)
-    , search: Nothing
-    , getItemCount: maybe 0 length <<< RD.toMaybe <<< _.available
-    , selections: []
-    , available: RD.NotAsked
+  ( dropdown :: D.Slot (Const Void) Unit )
```
thomashoneyman:
```suggestion
  ( dropdown :: D.Slot Unit )
```

examples/Components/Typeahead.purs
```diff
@@ -44,97 +38,90 @@ data Message
...
-            Just str, Nothing ->
-              Just (filter (_ /= mkLocation str) st.selections)
-            Just old, Just new ->
-              Just (mkLocation new : (filter (_ /= mkLocation old) st.selections))
-        for_ newSelections \selections ->
-          H.modify_ _ { selections = selections }
-
-  render :: S.State State -> H.ComponentHTML (S.Action Action) ChildSlots Aff
-  render st =
+       selections' <- Hooks.get selectionsId
```
thomashoneyman: It's OK to use `get` here, but it's not necessary. This could also just be `pure $ Just $ reply selections`.

spago.dhall
thomashoneyman: Sorry, that's right.

src/Select.purs
```diff
@@ -90,266 +77,304 @@ derive instance ordVisibility :: Ord Visibility
...
     where
-    -- we know that the getItemCount function will only touch user fields,
-    -- and that the state record contains *at least* the user fields, so
-    -- this saves us from a set of unnecessary record deletions / modifications
-    userState :: State st -> {| st }
-    userState = unsafeCoerce
-
-    lastIndex :: State st -> Int
-    lastIndex = (_ - 1) <<< st.getItemCount <<< userState
+      -- | An array of `IProps` with `ToggleProps`. It
```
thomashoneyman: These comments aren't going to show up in documentation, so we ought to add this documentation elsewhere, perhaps in a table in the README or something. That can be done in a subsequent update, just noting it here.

thomashoneyman (approved):
I'll leave this open for a couple days for @davezuch or other CitizenNet folks to weigh in if they want to, and then we can merge it.

src/Select.purs
JordanMartinez: Yeah... I was wondering about that...

examples/Components/Dropdown.purs
JordanMartinez: Done.

examples/Components/Typeahead.purs
JordanMartinez: Also done.

examples/Components/Typeahead.purs
JordanMartinez: Done.

spago.dhall
JordanMartinez: Done.

src/Select.purs
JordanMartinez: So... What if we put it on top of the type declarations (e.g. `ToggleProps`).

JordanMartinez: Or maybe on top of `SelectReturn`? Because more people are likely to hover over that...

bower.json
```diff
@@ -31,12 +31,14 @@
     "generated-docs"
   ],
   "dependencies": {
-    "purescript-halogen": "^5.0.0-rc.4",
-    "purescript-record": "^2.0.0"
+    "purescript-halogen": "^5.0.0-rc.9",
+    "purescript-halogen-hooks": "v0.4.0",
```
thomashoneyman: If these aren't in the bower registry and `bower install` would fail, can we point these at GitHub links instead?

src/Select.purs
thomashoneyman: I think over the type declarations makes the most sense

bower.json
JordanMartinez: Yeah. Done.

src/Select.purs
JordanMartinez: Done. I added `ContainerProps` even though it's not really needed so that we can be consistent.

JordanMartinez:
I think I've addressed everything. Phew!

PR #57, opened 2020-06-18
feat: update to next-6

srghma:

Please describe the purpose of the pull request, including any background context and links to related issues. If this introduces a large change to the project, please describe what you have implemented.

Help save the reviewer time by highlighting the most important sections of code.

If this is a new feature, you should provide new test coverage. If it fixes an existing bug, you should provide a failing test case this change fixes. Other than automated testing, what manual tests can the reviewer perform to verify your work?

## Other Notes:

Does the project documentation need to be updated? Should we introduce a new example, or update any existing examples? Does this PR require any other follow-up tasks?

### DIFF
--- a/src/Select.purs
+++ b/src/Select.purs
@@ -249,10 +249,10 @@ useSelect inputRec =
       -- | See `ToggleProps` for docs.
       toggleProps :: forall toggleProps. _ -> Array (HP.IProp (ToggleProps toggleProps) (HookM m Unit))
       toggleProps stateId =
-        [ HE.onFocus \_ -> Just (setVisibility stateId On)
-        , HE.onMouseDown \ev -> Just (toggleClick stateId ev)
-        , HE.onKeyDown \ev -> Just (key stateId ev)
-        , HE.onBlur \ev -> Just (setVisibility stateId Off)
+        [ HE.onFocus \_ -> setVisibility stateId On
+        , HE.onMouseDown \ev -> toggleClick stateId ev
+        , HE.onKeyDown \ev -> key stateId ev
+        , HE.onBlur \ev -> setVisibility stateId Off
         , HP.tabIndex 0
         , HP.ref (H.RefLabel "select-input")
         ]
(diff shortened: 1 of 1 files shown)


thomashoneyman:
@srgma Can you please fill out the PR template for future pull requests? It’s there so that the team here can get the gist of the changes before diving into the code. We can review your PRs and get them merged much faster if you help explain the motivation for changes and why they were done a certain way.

In this particular case you’re updating to Halogen 6, so you could just delete the issue template and say that.

I think it’s a bit premature to update given that Halogen 6 doesn’t have a release candidate yet and there may be more changes that require more updates to this library. Is there a particular reason you think the update should happen now?

If so you will also need to update the Spago files so the right version of Halogen is used. You’d need to override the package set to point at Halogen’s next-6 branch.

srghma:
@thomashoneyman updates

No, I dont want this pr to be merged now, I have created it just to test that it works with my hydration test project (that is rebased on next-6 as of now)

thomashoneyman:
I see. If you don’t want it merged, then can you use a fork of the library to test this instead of keeping a PR open here? Having the PR gives the sense that this is intended to be merged.

srghma:
> can you use a fork of the library to test this instead of keeping a PR open here?

I thought you will be interested

I would leave it open until next-6 is released (I plan to keep up to date with next-6 and may update this pr later when new changes are introduced)
I dont see a problem with lingering pr's

but yes, we can merge it to new branch next-6 or even close this

src/Select.purs
```diff
@@ -162,11 +161,17 @@ data SelectEvent
...
-derive instance newtypeUseSelect :: Newtype (UseSelect hooks) _
+type UseSelect' =
+  UseState SelectState
+    <> UseDebouncer String
+    <> UseEffect
+    <> Hooks.Pure
+
+instance newtypeUseDebouncer
+  :: HookEquals UseSelect' h
+  => HookNewtype UseSelect h
```
purefunctor: `hooks v0.5.0` drops `HookEquals` as its already possible to use type synonyms in this context:
```suggestion
  => HookNewtype UseSelect UseSelect'
```

PureFunctor (changes requested):
I tried migrating the hooks version of `halogen-formless` to Halogen 6 and had to do some minor edits for the examples to build:

